Issue #8, opened 2020-10-17
Create the MainActivity bottom navigation tabs


zhpeng811:
Create the bottom navigate buttons and the top tab
Investigate how fragments can be displayed inside the activity

please use the material design library, you can find some tutorial regarding the components [here](https://material.io/develop/android/components/bottom-navigation) 

please follow the [coding standard](https://www.oracle.com/java/technologies/javase/codeconventions-contents.html) and add javadoc comments where appropriate

Image from Mock UI:
![image](https://user-images.githubusercontent.com/43459469/96327370-17c3a000-0ff6-11eb-9bbc-ff1732c407c3.png)

PR #17, opened 2020-10-18
chore: create 4 bottom navs and their fragments+controllers

c25vdw:
part 1 of issue #8.

## new stuff
1. four packages under which there's their corresponding fragments + controllers (viewModel in this case.)
2. a new icon for the Profile button. can change later.
3. gradle update

### View Models
view models is a concept that's basically controllers but is closer to views than database. We should represent all UML controllers with view models. https://www.google.com/search?q=view+models


app/src/main/res/menu/bottom_nav_menu.xml
```diff
@@ -1,19 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <menu xmlns:android="http://schemas.android.com/apk/res/android">
-
+    <!-- icons subject to change  -->
     <item
-        android:id="@+id/navigation_home"
-        android:icon="@drawable/ic_home_black_24dp"
-        android:title="@string/title_home" />
+        android:id="@+id/navigation_library"
+        android:icon="@drawable/ic_home_black_24"
```
zhpeng811, line 6: Maybe change the name of all the drawable icons to match the corresponding fragment name? Everything else looks good for me

c25vdw: imo icons might be used more than one place, might not even be a fragment who's using it. It should then keep its name independent and only describe what it looks like. i.e. home icon is named home because it looks like a home instead of where it's used.

kdvuong: agreed with lucas

zhpeng811 (approved):

kdvuong:
No need to change for this PR but for future reference, I think all new stuffs added should have commit message "feat: ....".
"chores: ...." is for non-production code change.

kdvuong (approved):

Issue #7, opened 2020-10-17
Create the LoginActivity page


zhpeng811:
Implement the login activity page that allows a specific user to login

you can ignore the "Sign in with Google" part for now if it takes too long to understand the API and implement, this can be done as a separate low priority task later on

The "Create One" text which will redirect user to create an account can also be ignored for now, this will be done as a separate task, you can manually create a user document in firestore if you want to test the login functionality

please use the material design library to implement the textFields, the compoents you need are:
* com.google.android.material.textfield.TextInputLayout
* com.google.android.material.textfield.TextInputEditText
* you can find some tutorial regarding the components [here](https://material.io/develop/android/components/text-fields) 

please follow the [coding standard](https://www.oracle.com/java/technologies/javase/codeconventions-contents.html) and add javadoc comments where appropriate

**Note**: 
* This issue needs to be done after #6 or both issues need to be worked on concurrently

**Images**:
* Images can be directed exported from figma
* Feel free to use other images if you have better suggestions than the one we used in Mock UI

Image from Mock UI:
![image](https://user-images.githubusercontent.com/43459469/96325724-5e5dce00-0fe7-11eb-9139-909c27975b68.png)

PR #18, opened 2020-10-18
feat: #7:login activity feature

zhpeng811:
Implements #7 

Add a login page with username and password field to allow user to type their information to login


app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+     * android lifecycle method, grab the Layout and EditText fields
+     * as well as adding a simple textChangedListener
+     * @param savedInstanceState - the saved objects, should contain nothing for this activity
+     */
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_login);
+        usernameLayout = findViewById(R.id.login_username_layout);
+        usernameField = findViewById(R.id.login_username_field);
```
kdvuong, line 41: I believe the field is coupled with the layout right? I think you could do usernameField = usernameLayout.getEditText()

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+        usernameField.addTextChangedListener(new AfterTextChangedWatcher() {
+            @Override
+            public void afterTextChanged(Editable usernameString) {
+                String username = usernameString.toString();
+                checkUsername(username); // return value not needed
+            }
+        });
+        passwordLayout = findViewById(R.id.login_password_layout);
+        passwordField = findViewById(R.id.login_password_field);
```
kdvuong, line 50: same as above.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+import com.cmput301f20t21.bookfriends.R;
+import com.google.android.material.textfield.TextInputEditText;
+import com.google.android.material.textfield.TextInputLayout;
+
+/**
+ * Activity that will display the login page and handle user inputs
+ */
+public class LoginActivity extends AppCompatActivity {
+    private TextInputLayout usernameLayout;
+    private TextInputLayout passwordLayout;
```
kdvuong, line 27: We need an email field as well. Firebase only supports create user by email.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_login);
+        usernameLayout = findViewById(R.id.login_username_layout);
+        usernameField = findViewById(R.id.login_username_field);
+        usernameField.addTextChangedListener(new AfterTextChangedWatcher() {
+            @Override
+            public void afterTextChanged(Editable usernameString) {
+                String username = usernameString.toString();
+                checkUsername(username); // return value not needed
```
kdvuong, line 46: This will set an error when user type 1 character and delete. I don't think that's necessary. We can set error only once when login is clicked. We could make afterTextChanged setError to null only when user types.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * method called when user clicked the "login" button
+     * the function will validate the inputs and authenticate the user
+     * user will be redirected to the main screen if authentication is successful
+     * @param view - the view associated with the button
+     */
+    public void onLoginClicked(View view) {
+        Editable usernameText = usernameField.getText();
```
kdvuong, line 106: if usernameField = usernameLayout.getEditText(), I don't think usernameText will ever be null. Don't think we need to check it.

kdvuong (commented):
Some minor comments, the rest looks good to me.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
kdvuong: If we want to be really sure, we could check if getEditText() exists.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * android lifecycle method, grab the Layout and EditText fields
+     * as well as adding a simple textChangedListener
+     * @param savedInstanceState - the saved objects, should contain nothing for this activity
+     */
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_login);
```
c25vdw, line 39: 1. what if we [hide the action bar](https://stackoverflow.com/a/22338837/7358099) for login?
2. same for [the status bar](https://developer.android.com/training/system-ui/status#41)

kdvuong: I think the app flows like this.
User open apps -> login activity -> main activity.
Action bar and status bar should only exist beyond main activity

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
zhpeng811: getEditText() does not work for this, material design text field use TextInputEditText class instead of EditText class

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
zhpeng811: ken said he prefer non-email username, do we need email to do authentication? I think for now we can just use query for username and checking for password for authentication. Using other API requires more work(which can be a lower priority task later on)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
c25vdw: These codes under the `setContentView` should suffice to hide both status and action bar.
```
this.getSupportActionBar().hide();
View decorView = this.getWindow().getDecorView();
int uiOptions = View.SYSTEM_UI_FLAG_FULLSCREEN;
decorView.setSystemUiVisibility(uiOptions);
```

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
kdvuong: yeah we need email for authentication. Firebase has its own auth api which needs email. For username, we can make a new user name collection that has email as key and username as value.
Firebase already provides auth using email and password, doing secured password authentication ourselves will be redundant.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
kdvuong: I could do EditText usernameField = usernameLayout.getEditText() with no error locally.

zhpeng811: fixed, need a explicit cast to make it work

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
zhpeng811: fixed

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
zhpeng811: yeah before if I just used getText().toString() it gives a possible NPE warning, using getEditText().getText().toString() the warning disappeared, removed the check

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivity.java
zhpeng811: the hiding status bar part in the above code is deprecated in API 30, but I cannot find alternatives for it. Using it as for now, both action bars and status bars are now hidden for login activity

c25vdw:
Correct me if I'm mistaken, but I thought as discussed we will use this activity as the startup activity, and on finish start the main activity, instead of rendering the main activity on start up. Do we still do that in this PR?

zhpeng811:
> Correct me if I'm mistaken, but I thought as discussed we will use this activity as the startup activity, and on finish start the main activity, instead of rendering the main activity on start up. Do we still do that in this PR?

forgot about that after I made all the changes, l'll do it now

c25vdw (approved):
Looks good to me.

kdvuong (approved):
looks good to me

PR #19, opened 2020-10-19
chore: add firestore dependency

kdvuong:
Note: downgrading from gradle 6.5 to 6.1.1.
Issue link: https://github.com/gradle/gradle/issues/14889


kdvuong:
This seems to work for me locally. You guys should try fetching this branch and build see if it works.

zhpeng811 (approved):

zhpeng811:
I didn't encounter any issue building locally

c25vdw (approved):
Everyone should fetch this and try to build to make sure it's not broken. Otherwise I'm good with downgrade if there's no immediate fix for this. Maybe we should be attentive to if the issue disappeared (by when we should all update Gradle)

c25vdw:
merging. 
in case the app doesn't build, do a manual gradle sync https://stackoverflow.com/a/29565362/7358099

PR #27, opened 2020-10-21
refactor: use lambda callbacks for LoginController handler

c25vdw:


kdvuong:
I think this is a `refactor: ...` not `chore:...`

app/src/main/java/com/cmput301f20t21/bookfriends/controller/LogInController.java
```diff
@@ -14,22 +14,27 @@
...
 
-    public LogInController(ILogInListener listener) {
+    public LogInController() {
         authService = AuthService.getInstance();
         userService = UserService.getInstance();
-        this.listener = listener;
     }
 
-    public void handleLogIn(final String username, final String password) {
+    public void handleLogIn(final String username, final String password, final OnSuccessCallback callbackSuccess, final OnFailCallback callbackFailure) {
```
kdvuong, line 37: rename to `successCallback` instead of `callbackSuccess`? same for failure

zhpeng811 (approved):
looks good to me

one note is the error to display can just be a general one like "username or password is incorrect", but this can be done in the integration task.

kdvuong (approved):
lgtm

PR #30, opened 2020-10-21
feat: add Book entity and BookService

c25vdw:
not yet fleshed out. can we discuss and keep working on this?

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
@@ -1,0 +1,31 @@
+package com.cmput301f20t21.bookfriends.services;
+
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.firestore.CollectionReference;
+import com.google.firebase.firestore.FirebaseFirestore;
+import com.google.firebase.firestore.QuerySnapshot;
+
+public class BookService {
+    // TODO not done
+
+    private CollectionReference bookCollection;
+
+    private static final BookService instance = new BookService();
+
+    private BookService() {
+        bookCollection = FirebaseFirestore.getInstance().collection("books");
+    }
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private BookService() {
+        bookCollection = FirebaseFirestore.getInstance().collection("books");
+    }
+
+    public static BookService getInstance() {
+        return instance;
+    }
+
+    public Task<QuerySnapshot> getBooksOfOwnerID(String uid) {
```
kdvuong, line 23: This can be combine with the one below to be getBooksByUID?

kdvuong:
yep. I think having getBooks and addBook is enough for this PR. Everything else will be added whenever we need in future PR.

app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
c25vdw: how do we figure out the books that's borrowed by this user?
Option 1: have a borrower field in Book
Option 2: query for the user's requests whose book is borrowed.

kdvuong: I think option 2 is better

QuocTrungTran: I prefer option 2 as well

c25vdw: We will need the Request entity for that functionality. In another PR, maybe?

kdvuong (approved):
lgtm

zhpeng811 (approved):
looks good for this PR

we need to add the optional description field in another PR

PR #63, opened 2020-10-22
feat: store username in auth service

kdvuong:
allow us to get username after log in by doing: `AuthService.getInstance().getCurrentUser().getUsername();`

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/entities/User.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/entities/User.java
@@ -1,21 +1,25 @@
 package com.cmput301f20t21.bookfriends.entities;
 
 public class User {
-    private String id;
+    private String uid;
+    private String username;
     private String displayName;
     private String email;
     private String phoneNumber;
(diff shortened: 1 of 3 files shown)


zhpeng811:
question: what is displayedName? Original I thought that is the username but now username have its own field

kdvuong:
> question: what is displayedName? Original I thought that is the username but now username have its own field

Display name could be Khang Vuong. Doesn't have to be unique. We can drop it if we dont need it. Username could be khangvuong and has to be unique.

zhpeng811:
> > question: what is displayedName? Original I thought that is the username but now username have its own field
> 
> Display name could be Khang Vuong. Doesn't have to be unique. We can drop it if we dont need it. Username could be khangvuong and has to be unique.

I think we don't need it, currently there's nowhere that the user can set their displayed name. Your call if it needs to be removed in this PR or later on, otherwise it looks good!

zhpeng811 (approved):

c25vdw (approved):
lgtm

PR #66, opened 2020-10-23
Bug fixed

Meilin-Lyu:
all errors will be shown at the same time once clicking "create"
latest check will overwrite the error message


QuocTrungTran (approved):
LGTM! 
But I think we should start adding javadoc as we code. So we don't have to go back to add them later.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -93,18 +93,85 @@ private boolean checkEmpty(TextInputLayout layout) {
...
+
+        if (password.length() < 6) {
+            passwordLayout.setError(getString(R.string.password_length_error));
+            isIncorrectLength = true;
+        }
+
+        return isIncorrectLength;
+    }
+
+    private boolean isEmailValid() {
```
c25vdw, line 133: suggestion: rename to `checkEmailFormat` for consistency

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -93,18 +93,85 @@ private boolean checkEmpty(TextInputLayout layout) {
...
+        boolean checkPasswordLength = checkPasswordLength();
+        boolean checkUsername = checkUsername();
+        boolean checkUsernameSpace = checkSpace(usernameLayout);
+        boolean checkPasswordSpace = checkSpace(passwordLayout);
+        boolean checkUsernameEmpty = checkEmpty(usernameLayout);
+        boolean checkConfirmEmpty = checkEmpty(confirmLayout);
+        boolean checkPasswordEmpty = checkEmpty(passwordLayout);
+        boolean checkEmailEmpty = checkEmpty(emailLayout);
+        boolean checkPasswordMatch = checkPasswordMatch();
+        if (checkEmailValid&&!checkPasswordLength&&!checkUsername&&!checkUsernameSpace
```
c25vdw, line 172: The logic here looks messy to me because we have some checkers that returns true to be valid and some that's false to be valid (like `notAlphaNumeric` for checkUsername) 

suggestion: we need to inverse the logic and function name for them. i.e. checkUsername -> checkUsernameValid, and returns isAlphaNumeric. same for checkEmpty -> checkNotEmpty, etc. that way the if logic here will be very straight forward instead of seeing a lot of `!` logic

kdvuong: Rather than name checkUsernameValid, I think isUsernameValid would be a better name. Check should be the function name and it returns isValid.

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.1'
+        classpath 'com.android.tools.build:gradle:4.1.0'
```
kdvuong, line 8: this will break sync. Should revert backk to 4.0.1

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,6 @@
-#Sun Oct 18 13:09:01 MDT 2020
+#Thu Oct 22 19:36:02 MDT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-bin.zip
-
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-bin.zip
```
kdvuong, line 6: same here, should revert back to 6.1.1

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -93,18 +93,85 @@ private boolean checkEmpty(TextInputLayout layout) {
...
+        boolean hasSpace = false;
+        if (userInput.contains(" ")) {
+            layout.setError(getString(R.string.space_error));
+            hasSpace = true;
+        }
+
+        return hasSpace;
+    }
+
+    private boolean checkUsername() {
```
kdvuong, line 108: I think all these validator functions can go into view model so we can keep our activity class clean. When moving to view model, we can make the function purer by only take in a String param and return boolean. Setting error can still be handled by the activity class.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -77,58 +81,104 @@ public void afterTextChanged(Editable emailString) {
...
+     * @param view
+     */
     public void onCreateClicked(View view){
-        String confirmPassword = confirmField.getText().toString();
-        String password = passwordField.getText().toString();
-        if(!confirmPassword.equals(password)){
+        boolean isEmailValid = model.isEmailValid(emailLayout);
+        boolean correctPasswordLength = model.correctPasswordLength(passwordLayout);
+        boolean isUsernameValid = model.isUsernameValid(usernameLayout);
+        boolean usernameNoSpace = model.noSpace(usernameLayout);
```
kdvuong, line 99: I think its redundant that we keep negate no space and notEmpty. We could check for is empty like
`userNameIsEmpty = model.isEmpty(usernameLayout)`

kdvuong: user space can also be checked in alpha numeric? Please verify this, I'm not too sure.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -77,58 +81,104 @@ public void afterTextChanged(Editable emailString) {
...
+        boolean passwordNoSpace = model.noSpace(passwordLayout);
+        boolean usernameNotEmpty = model.notEmpty(usernameLayout);
+        boolean confirmNotEmpty = model.notEmpty(confirmLayout);
+        boolean passwordNotEmpty = model.notEmpty(passwordLayout);
+        boolean emailNotEmpty = model.notEmpty(emailLayout);
+        boolean isPasswordMatch = model.isPasswordMatch( passwordLayout, confirmLayout);
```
kdvuong, line 105: I think we don't have to declare all these variable since model functions will be descriptive enough.
We can just do: `if (model.isPasswordMatch( passwordLayout, confirmLayout)) {...}
Same for all above.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/login/CreateAccountActivity.java
```diff
@@ -77,58 +81,104 @@ public void afterTextChanged(Editable emailString) {
...
+        }
+        if(!isUsernameValid){
+            usernameLayout.setError(getString(R.string.alphanumeric_error));
+            allValid = false;
+        }
+        if(!usernameNoSpace){
+            usernameLayout.setError(getString(R.string.space_error));
+            allValid = false;
+        }
+        if(!passwordNoSpace){
```
kdvuong, line 123: I did a quick google search and seems like its normal for password to have space so no need for check here.
https://stackoverflow.com/questions/632167/should-users-be-allowed-to-entered-a-password-with-a-space-at-the-beginning-or-e#:~:text=Space%20is%20a%20regular%20password,treated%20as%20any%20other%20character.

zhpeng811 (approved):

kdvuong (approved):
lgtm

PR #67, opened 2020-10-23
fix: fix uid and username order in AuthService

zhpeng811:
A small bug here that AuthService was creating a User class using `username` as first parameter and `uid` as second parameter, but the User class constructor have the reversed order

so before if I call authService.getInstance().getCurrentUser().getUsername() it will return me the `uid` instead


kdvuong:
oops :)

kdvuong (approved):
lgtm

c25vdw (approved):
lgtm

Meilin-Lyu (approved):
lgtm

PR #68, opened 2020-10-24
feat: integrate add book to firestore along with the image

zhpeng811:
* Adding a book will now store the book information to Firestore
* If the book have an image attached with it, the image will also be stored to Firebase cloud storage, the name of the image will be the id of the book 
* some commented code in AddEditActivity gives an idea on how we can retrieve the image from cloud


zhpeng811:
one note: we might want to have general callback interfaces for all ViewModels, as now each ViewModel is having duplicate callback interfaces, and will continue to grow as we add more ViewModels. 

But some callbacks requires parameter(s) while some does not. If you have better suggestions maybe we can discuss.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /** called by handlers when async request failed */
+    public interface OnFailCallback {
+        void run(BOOK_ERROR error);
+    }
+
+    private final AuthService authService = AuthService.getInstance();
+    private final BookService bookService = BookService.getInstance();
+    private final StorageReference imageStorageReference = FirebaseStorage.getInstance().getReference();
```
kdvuong, line 37: I think since photo is related to book, we can hide this info in the bookService as well.

kdvuong:
> one note: we might want to have general callback interfaces for all ViewModels, as now each ViewModel is having duplicate callback interfaces, and will continue to grow as we add more ViewModels.
> 
> But some callbacks requires parameter(s) while some does not. If you have better suggestions maybe we can discuss.

I have thought about this too. We could create 3 interfaces, onSuccessCallback, onSuccessCallbackWithArgs, and onFailureCallback. onSuccessCallback and onFailureCallback we already have. onSuccessCallbackWithArgs will take in a Map<String, Object>. We can send our arguments through callback using the map. We can extract the data in the activity after the callback is called. This approach makes it less clear on what the activity is expecting from the callback, open to suggestions.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/LibraryFragment.java
```diff
@@ -31,20 +32,19 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
         return root;
     }
 
     /**
      * function allows user to jump into the add/edit screen when click on the floating button
      */
     private void openAddEditActivity() {
+        // TODO: Change the enum when calling the activity for editing
         Intent intent = new Intent(this.getActivity(), AddEditActivity.class);
+        intent.putExtra("BOOK_ACTION", BOOK_ACTION.ADD);
```
QuocTrungTran, line 47: I think a good practice here is to declare the key "BOOK_ACTION" as static final. Something like this:
public static final String BOOK_ACTION_KEY = "com.cmput301f20t21.bookfriends.BOOK_ACTION";
And then in AddEditActivity.java:
action = (BOOK_ACTION) intent.getSerializableExtra(LibraryFragment.BOOK_ACTION_KEY);

https://stackoverflow.com/questions/57010358/best-practice-for-name-putextra/57010913

QuocTrungTran (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/LibraryFragment.java
```diff
@@ -20,6 +20,8 @@
 
 public class LibraryFragment extends Fragment {
 
+    public static final String BOOK_ACTION_KEY = "com.cmput301f20t21.bookfriends.BOOK_ACTION";
```
c25vdw, line 23: we might need to put this into the BOOK_ACTION enum file. Or, with
```java
BOOK_ACTION.class.getName()
``` 
you get the same result

c25vdw: sorry didn't notice trung's comment. I'm good with both solution.

kdvuong (approved):
lgtm

c25vdw (approved):
lgtm

c25vdw:
agree with [khang's comment](https://github.com/CMPUT301F20T21/Book_Friends/pull/68#issuecomment-715680670) but a simple generic typed message should suffice? view models could use their own enum message rather than a map.

have a example commit at https://github.com/CMPUT301F20T21/Book_Friends/commit/6176183a3c18a618e02ac9e55f5902f2ec956bad

PR #70, opened 2020-10-24
feat: add book list view for library fragment

c25vdw:
what the classes do:

- LibraryTabCollectionAdapter: the adapter that maps tab to which fragment to show under the tab.

for each of the owner and borrower tab, there's:
- xxxListAdapter: the recycler list adapter that maps a `Book` to a book item view
- xxxListFragment: the view below the top tab. right now just a recycler but more stuff to come, i.e. filter button etc.
- xxxViewModel: the vm that serves for the list fragment. controls what data to show.

Further work related to integration:
- fragment/adapter need to update upon view model data changes.
- book item menu will change, both view and functionality.


c25vdw:
demo
output5

kdvuong:
`BorrowedListAdapter`` and `OwnedListAdapter` are really similar. I think we can make an abstract `BaseBookListAdapter` and have the 2 inherit it? All we need to do is pass in the `R.layout.id` for `onCreateViewHolder` to inflate. We can add the moreButton in `onCreateViewHolder` as well?

c25vdw:
UX issue found :(
image

c25vdw:
<details><summary>expand to see updated demo, avoid button overlapping</summary>
</img></details>

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static class ViewHolder extends RecyclerView.ViewHolder {
+        TextView title;
+        TextView author;
+        TextView owner;
+        TextView isbn;
+        ImageButton moreBtn;
+
+        Book book;
+        View root;
```
zhpeng811, line 57: question: is there a reason that these attributes are all default?

c25vdw: https://stackoverflow.com/a/215505/7358099
had no idea about access control when creating them. but looking back on this, I should make them private and extract a binding method for the view holder. For right now, it works, and default is more restricted than public. will make a change.

c25vdw: actually I think we should stick to default. I am currently working on abstracting the list adapter and the view holder. having child view holder inheriting the parents' views means that those parent view needs to be at least default. for children to access it, default is the best option imo.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BaseBookListAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public int getItemCount() {
+        return books.size();
+    }
+
+    public static class ViewHolder extends RecyclerView.ViewHolder {
+        // TODO flesh out book view card
+        final TextView title;
+        final TextView owner;
```
kdvuong, line 43: we should remove owner field here. BorrowedBook should handle displaying the owner instead of having OwnedBook hide it.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedListAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+            owner = v.findViewById(R.id.item_book_owner);
+            moreBtn = v.findViewById(R.id.item_book_more_btn);
+            moreBtn.setOnClickListener(this::showPopup);
+        }
+
+        private void showPopup(View view) {
+            PopupMenu popup = new PopupMenu(view.getContext(), moreBtn);
+            MenuInflater inflater = popup.getMenuInflater();
+            inflater.inflate(R.menu.library_borrowed_book_item_menu, popup.getMenu());
+            popup.setGravity(Gravity.END);
```
kdvuong, line 47: Do we want the moreButton on top or at the bottom? I think either is fine but I prefer on top since most app have it on top. Bottom can be used to display status.

c25vdw: let's do on top.

zhpeng811 (approved):
lgtm

kdvuong (approved):
lgtm

PR #71, opened 2020-10-25
feat: add request service

kdvuong:


app/src/main/java/com/cmput301f20t21/bookfriends/services/RequestService.java
```diff
@@ -0,0 +1,95 @@
...
+        return batch.commit();
+    }
+
+    /**
+     * get request that unavailable.
+     * unavailable requests have status: ACCEPTED, BORROWED
+     * @param bookId
+     * @return Task QuerySnapshot for unavailable request
+     */
+    public Task<QuerySnapshot> getUnavailableRequest(String bookId) {
```
kdvuong, line 89: I'm thinking when we get book, we can fill in the book status by check for requests that are accepted or borrowed. If this query return empty, then book is available. If we implement the system correctly, this query should return only 1 document. We can set the book status to the request status.

c25vdw: This could work. But in comparison, what if we store book status as a Book attribute, update them as we update requests if necessary? if we did it right, it should save a lot of query time, and separate book status from requests'.

consider when we fetch all borrowed books: the method you mentioned might query in time (#of requests  * # of books )

c25vdw:
question: will the Request class be part of this PR? other than that LGTM!

app/src/main/java/com/cmput301f20t21/bookfriends/services/RequestService.java
zhpeng811: by US 01.03.01:
As an owner or borrower, I want a book to have a status to be one of: available, requested, accepted, or borrowed.

I think book does need to have a status attribute

app/src/main/java/com/cmput301f20t21/bookfriends/entities/Request.java
```diff
@@ -0,0 +1,33 @@
+package com.cmput301f20t21.bookfriends.entities;
+
+import com.cmput301f20t21.bookfriends.enums.REQUEST_STATUS;
+
+public class Request {
+    private String id;
+    private String requesterId;
+    private String bookId;
+    private REQUEST_STATUS status;
```
c25vdw, line 9: do we want to have the owner user id? I know this can be found in the book entity, but it would be better if we have that attribute flatten out here, which is also needed/faster when we query for "incoming requests".

kdvuong: addressed in meeting.

zhpeng811 (approved):
lgtm

c25vdw (approved):
lgtm

PR #76, opened 2020-10-28
feat: Integrate owner library

zhpeng811:
Connect "My Books" tab to
* firestore to fetch a list of books for the current user upon view created
* firebase cloud storage to fetch the image(if exist) for each book

Features not in this PR:
* Edit and Delete a book
* Update the list dynamically after user add a book (Currently there is a bug that after adding a book the entire view will be pushed up, not sure why that happened and I will demonstrate in the meeting)

<details><summary>DEMO</summary>
<p>

screencast-nimbus-capture-2020 10 28-13_23_41

</p>
</details>


app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
```diff
@@ -29,13 +34,16 @@ public static BookService getInstance() {
...
+    public Task<DocumentReference> add(String isbn, String title, String author, String description, String owner, boolean imageAttached) {
+        HashMap<String, Object> data = new HashMap<>();
         data.put("isbn", isbn);
         data.put("title", title);
         data.put("author", author);
         data.put("description", description);
         data.put("owner", owner);
+        // when a book is first added, the status will always be "AVAILABLE"
+        data.put("status", BOOK_STATUS.AVAILABLE.toString());
+        data.put("image", imageAttached);
```
zhpeng811, line 46: When I tried to get a image that does not exist in firebase cloud storage, the API is throwing exception in its own class that onCompleteListener can't even catch. 

The solution I used here is to add a boolean variable to the book collection indicating if this book have an image attached with it. So the ViewModel will only attempt to get image that does exist in the cloud storage, this solution can also reduce the number of calls to firebase.

If you have better solution we can discuss

kdvuong:
Found a bug, when owner has no book, add button appear on top
create

app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
c25vdw: on that, did you try adding the onfail listener for the task? checkout https://stackoverflow.com/a/57446232/7358099

kdvuong: @c25vdw I can confirm what @zhpeng811 said, I added on fail listener and error is handled but the error still appear in the log.

c25vdw: If the exception is properly handled why would we have a boolean indicator on whether we should fetch? imo it's a larger cost to add an attribute in Book specific for this action.

kdvuong: It's a trade-off. With @zhpeng811 solution, the log looks cleaner but clutter the Book document. I think we can remove the boolean from Book if we can properly handle the exception.

c25vdw: agree, it is a tradeoff. And I think we shouldn't care about whatever the log spits out if it does not crash the app. Also we should keep Book entity free from any app behaviours.

zhpeng811: It depends on what we want, the error still appears in the log so I'm not certain to say that it's properly handled, on my testing it does not break anything as for now but there's no promise if it won't break anything in the future.

With the current solution, we can reduce the number of getDownloadURL() calls to (# of books with images), where as without the boolean would be (# of books)

zhpeng811: image

here is the error, it will display once for every book that does not have an image (so if owner have 10 books and 6 books without image it will display 6 times)

zhpeng811:
> Found a bug, when owner has no book, add button appear on top
> 
> Demo

found a fix and opening another PR with the change

app/src/main/java/com/cmput301f20t21/bookfriends/services/BookService.java
zhpeng811: edited to include and use imageName attribute

c25vdw (approved):
lgtm

kdvuong (approved):
Lgtm

PR #77, opened 2020-10-28
fix: use relative layout to stick FAB position to bottom of layout

zhpeng811:
This is the only solution I found [online](https://stackoverflow.com/questions/49998087/floating-action-button-is-also-scrolling-along-with-recyclerview) that actually works

<details><summary>Emulator screenshot</summary>
<p>

image
image

</p>
</details>

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
@@ -9,14 +9,7 @@ import java.util.ArrayList;
 public class OwnedViewModel extends ViewModel {
     public ArrayList<Book> getBooks() {
         ArrayList<Book> mData = new ArrayList<>();
-        mData.add(new Book ("The two towers" ));
-        mData.add(new Book ("The Lord of the Rings"));
-        mData.add(new Book ("Notes for CMPUT301"));
-        mData.add(new Book ("The Return of the King"));
-        mData.add(new Book ("The Hobbit"));
-        mData.add(new Book ("The Return of the King"));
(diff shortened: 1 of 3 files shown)


c25vdw:
instead of this PR, here's what i think is the right way to do this: 

1. add a constraint bottom2bottom in fragment_library/viewpager2 https://github.com/CMPUT301F20T21/Book_Friends/blob/5efececf44f11a64b614ee61dcddd14955c26f23/app/src/main/res/layout/fragment_library.xml#L20
2. remove the 48dp padding bottom in both constraint layouts (owned/borrowed) https://github.com/CMPUT301F20T21/Book_Friends/blob/5efececf44f11a64b614ee61dcddd14955c26f23/app/src/main/res/layout/owned_list_book.xml#L6

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
```diff
@@ -9,14 +9,7 @@
...
-        mData.add(new Book ("The two towers" ));
-        mData.add(new Book ("The Lord of the Rings"));
-        mData.add(new Book ("Notes for CMPUT301"));
-        mData.add(new Book ("The Return of the King"));
-        mData.add(new Book ("The Hobbit"));
-        mData.add(new Book ("The Return of the King"));
-        mData.add(new Book ("The Return of the King"));
-        mData.add(new Book ("The Return of theKing"));
+
         return mData;
```
c25vdw, line 13: we don't need changes in this file as viewmodel is being implemented

c25vdw (approved):

kdvuong (approved):

PR #78, opened 2020-10-29
Profile

Kii-26:


app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileFragment.java
```diff
@@ -47,6 +56,30 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
         editProfile = view.findViewById(R.id.image_edit);
+        name = view.findViewById(R.id.username);
+
+        // get the login information from firebase
+        User firebaseUser = AuthService.getInstance().getCurrentUser();
+        if (firebaseUser != null) {
+            String userId = firebaseUser.getUid();
+
+            //document reference
+            final DocumentReference docReference = FirebaseFirestore.getInstance().collection("users").document(userId);
```
zhpeng811, line 67: the User object firebaseUser already contains the information (username, email) that we need, you can get it by calling getUsername() and getEmail(). 

There's no need to get it from firestore again

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileEditDialog.java
```diff
@@ -35,8 +33,10 @@ public View onCreateView(final LayoutInflater inflater, @Nullable ViewGroup cont
         editPhone = view.findViewById(R.id.edit_phone);
         cancel = view.findViewById(R.id.text_cancel);
         confirm = view.findViewById(R.id.text_confirm);
-        existEmail = view.findViewById(R.id.email);
-        existPhone = view.findViewById(R.id.phone);
+        //set value from the fragment
+        editEmail.setText(getArguments().getString("email"));
```
zhpeng811, line 37: we should first get the arguments by doing `Bundle arguments = getArguments()` and check if it is null before proceeding to get the Strings

without doing this check there is a possibility that the arguments did not get passed in to the fragment and that will produce a NPE(Null Pointer Exception) and cause our app to crash

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileEditDialog.java
```diff
@@ -56,6 +71,28 @@ public void onClick(View v) {
...
+                FirebaseUser user = FirebaseAuth.getInstance().getCurrentUser();
+                user.updateEmail(inputEmail)
+                        .addOnCompleteListener(new OnCompleteListener<Void>() {
+                            @Override
+                            public void onComplete(@NonNull Task<Void> task) {
+                                if (task.isSuccessful()) {
+                                    Log.d(TAG, "User email address updated.");
+                                }
+                            }
+                        });
```
c25vdw, line 84: suggestion: extract this to a new method inside the [profile view model](https://github.com/CMPUT301F20T21/Book_Friends/blob/master/app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileViewModel.java) called `updateCurrentUserEmail`. use that function to do all the thing about updating emails (including the code below)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileEditDialog.java
```diff
@@ -56,6 +71,28 @@ public void onClick(View v) {
...
+                //update field "email" and "phone"
+                User firebaseUser = AuthService.getInstance().getCurrentUser();
+                if (firebaseUser != null) {
+                    String userId = firebaseUser.getUid();
+                    FirebaseFirestore.getInstance().collection("users").document(userId)
+                            .update(
+                                    "email", inputEmail,
+                                    "phone", inputPhone
+                            );
```
c25vdw, line 93: suggestion: after we moved the whole firebase-related section (line 74 ~ 93) to view model, I think we should still extract this part to UserService as a new method called `updateFirestoreUser`, which takes a HashMap<String, Object> and do the update call there.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileFragment.java
```diff
@@ -47,6 +63,25 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
+            String username = firebaseUser.getUsername();
+            String email = firebaseUser.getEmail();
+            name.setText(username);
+            emailAddress.setText(email);
+
+            //once the profile is opened, user can edit their phone number, and new field in firebase is created
+            HashMap<String, Object> data = new HashMap<>();
+            data.put("phone", phoneNumber.getText().toString());
+            FirebaseFirestore.getInstance().collection("users").document(userId)
+                    .set(data, SetOptions.merge());
```
c25vdw, line 82: I think this part of code is duplicate of the update logic in EditFragment (commented above) and should be removed. please double check on this part and maybe pick one.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileEditDialog.java
kdvuong: we might also want to abstract updateEmail in AuthService. We don't want to expose FirebaseUser class throughout the app because it has too much power (example: we could delete user by doing user.delete()).

Kii-26: For updateEmail in AuthService, so you mean I need to implement a method in updateEmail in User entity?

zhpeng811: I think Khang means move this section to AuthService class and have a new method called updateEmail(). Anything directly connecting to firebase/firestore should be done in the service classes.

app/src/main/java/com/cmput301f20t21/bookfriends/services/AuthService.java
```diff
@@ -43,4 +49,21 @@ public User getCurrentUser() {
     public void signOut() {
         mAuth.signOut();
     }
+
+    public void updateEmail(String username, String TAG){
+        FirebaseUser user = mAuth.getCurrentUser();
+        if (user != null) {
+            user.updateEmail(username)
+                    .addOnCompleteListener(new OnCompleteListener<Void>() {
```
zhpeng811, line 57: the OnCompleteHandler should exist in ViewModel instead of service, because currently there's no way for the fragment to know if the update succeed or failed(Log.d is only for developer debugging purposes and user cannot see that) 

the service should just do `return user.updateEmail(...)`, the ViewModel should have an instance of the service and call `authService.updateEmail(...)` then add handlers that can be directly pass back to the fragment.

Also I think the parameter should be `email` instead of `username` ? I don't see the email that is being updated here

If you are unsure how to proceed we can discuss during tonight's meeting and show an example from another fragment

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileFragment.java
```diff
@@ -47,6 +63,45 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
+        User firebaseUser = AuthService.getInstance().getCurrentUser();
+        if (firebaseUser != null) {
+            String userId = firebaseUser.getUid();
+            String username = firebaseUser.getUsername();
+            String email = firebaseUser.getEmail();
+            name.setText(username);
+            emailAddress.setText(email);
+
+            // search for field "phone"
+            final DocumentReference docReference = FirebaseFirestore.getInstance().collection("users").document(userId);
```
zhpeng811, line 79: this is directly connect to firebase as well and need to be in AuthService

Did we discuss that we might not need a phone number field? Let's bring it up in tonight's meeting and discuss

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileViewModel.java
```diff
@@ -75,4 +84,33 @@ public void getUserByUid(String uid, OnSuccessCallbackWithMessage<User> onSucces
...
+                        }
+                    });
+
+        }
+    }*/
+    public void updateFirestoreUser(String inputEmail, String inputPhone){
+        User firebaseUser = AuthService.getInstance().getCurrentUser();
+        if (firebaseUser != null) {
+            String userId = firebaseUser.getUid();
+            FirebaseFirestore.getInstance().collection("users").document(userId)
```
zhpeng811, line 109: similar to the above comment regarding email, this should be in UserService class, and have a OnCompleteHandler attached to that method instead

app/src/main/java/com/cmput301f20t21/bookfriends/services/UserService.java
```diff
@@ -49,4 +48,13 @@ public static UserService getInstance() {
                 .whereLessThan("username", username.concat("~")) // ascii ~ is way larger than 'z'
                 .limit(10).get();
     }
+
+    public Task<Void> updateUser(String email){
```
c25vdw, line 52: suggestion: refactor/rename to `updateUserEmail` as we are only updating email here.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/profile/ProfileViewModel.java
```diff
@@ -75,4 +85,31 @@ public void getUserByUid(String uid, OnSuccessCallbackWithMessage<User> onSucces
...
+                        @Override
+                        public void onComplete(@NonNull Task<Void> task) {
+                            if (task.isSuccessful()) {
+                                Log.d(TAG, "User email address updated.");
+                            }
+                        }
+                    });
+    }
+
+    public void updateFirestoreUser(String inputEmail, String TAG){
```
c25vdw, line 102: suggestion: refactor/rename to `updateFirestoreUserEmail`

c25vdw (approved):
lgtm

zhpeng811 (approved):
lgtm

PR #79, opened 2020-10-29
feat: integrate and implement BorrowedViewModel

kdvuong:
- I branched this off @zhpeng811 owner integration branch because I needed some of the stuff

<details><summary>DEMO</summary>

<p>

create

</p>

</details>


app/build.gradle
```diff
@@ -7,7 +7,7 @@ android {
 
     defaultConfig {
         applicationId "com.cmput301f20t21.bookfriends"
-        minSdkVersion 16
+        minSdkVersion 24
```
kdvuong, line 10: increase minimum api to use higher order function for list (like map)

kdvuong:
My changes are in BorrowedViewModel.java and BorrowedListFragment.java only.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedViewModel.java
```diff
@@ -1,22 +1,96 @@
...
+                    String title = (String) document.get("title");
+                    String author = (String) document.get("author");
+                    String description = (String) document.get("description");
+                    String owner = (String) document.get("owner");
+                    String status = (String) document.get("status");
+                    String imageName = (String) document.get("imageName");
+                    Book book = new Book(id, isbn, title, author, description, owner, imageName, BOOK_STATUS.valueOf(status));
+
+                    if (document.get("imageName") != null) {
+                        bookService.getImage((String) document.get("imageName")).addOnSuccessListener(uri -> {
```
kdvuong, line 83: with this approach, we're getting the image uri right away, no need to pass imageName to book constructor.
Once the uri is fetched succesfully, we will notify the adapter through the live data.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedViewModel.java
```diff
@@ -1,22 +1,96 @@
...
+                    DocumentSnapshot document = documents.get(i);
+                    String id = document.getId();
+                    String isbn = (String) document.get("isbn");
+                    String title = (String) document.get("title");
+                    String author = (String) document.get("author");
+                    String description = (String) document.get("description");
+                    String owner = (String) document.get("owner");
+                    String status = (String) document.get("status");
+                    String imageName = (String) document.get("imageName");
+                    Book book = new Book(id, isbn, title, author, description, owner, imageName, BOOK_STATUS.valueOf(status));
```
kdvuong, line 80: This whole block could be refactored

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedViewModel.java
zhpeng811: looks better this way with LiveData, it could also get rid of all the callbacks. Since The Owner section need to use LiveData anyway I'll refactor that branch now to match it

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedViewModel.java
zhpeng811: I can refactor this method in BookService to return a Book object for a single DocumentSnapshot
image

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedListFragment.java
```diff
@@ -41,8 +45,16 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         recyclerView.setLayoutManager(layoutManager);
 
         // specify an adapter (see also next example)
-        mAdapter = new BorrowedListAdapter(mViewModel.getBooks());
-        recyclerView.setAdapter(mAdapter);
+        vm.getBooks().observe(getViewLifecycleOwner(), (List<Book> books) -> {
+            adapter = new BorrowedListAdapter(books);
```
zhpeng811, line 49: one little thing there if you don't specify an adapter directly in the onViewCreated() it will thrown an error
image

Not sure if this error will affect anything, but a simple way to get rid of the error is to feed in a temporary adapter, once we get the data we will set it to the real adapter

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/BorrowedViewModel.java
```diff
@@ -1,22 +1,96 @@
...
-        mData.add(new Book ("Calculus IV"));
-        mData.add(new Book ("Calculator V"));
-        mData.add(new Book ("Terminator V"));
-        return mData;
+    private RequestService requestService = RequestService.getInstance();
+    private BookService bookService = BookService.getInstance();
+    private AuthService authService = AuthService.getInstance();
+
+    private MutableLiveData<List<Book>> books;
+    private MutableLiveData<Integer> updatedPosition;
```
c25vdw, line 30: on @zhpeng811 's comment about adapter not initialized, that's a result of the book list mutable live data not initialized. we can init them on view model creation, i.e. 
```java
class BorrowedVM {
    private final MutableLiveData<...> = new MutableLiveData<>(new ArrayList<>());
}

```
(by final we make sure it's not re-initialized elsewhere via `= new MutableLiveData(...);` or we lost the reactivity )

just an option to counter the issue

c25vdw (approved):
lgtm

zhpeng811 (approved):
lgtm

PR #81, opened 2020-10-30
feat: create base/basic scanner activity

c25vdw:
- create a scanner activity with its own layout that provides any extending children activity the functionality to scan barcodes. Any class that extends this class should take care of what to do with the scanned barcode.
- create a bare bone Scanner Add activity that works specifically for adding a book from scanned isbn code.

### what's not included
- actual adding book functionality from received isbn barcode, e.g. the data communication between add-edit and the scanner-add
- any extra data fetching upon successful scan (e.g. book info grabbed from internet)

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
@@ -19,12 +19,12 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.lifecycle.ViewModelProvider;
 
-import com.bumptech.glide.Glide;
 import com.cmput301f20t21.bookfriends.R;
 import com.cmput301f20t21.bookfriends.enums.BOOK_ACTION;
 import com.cmput301f20t21.bookfriends.enums.BOOK_ERROR;
 import com.cmput301f20t21.bookfriends.ui.library.OwnedListFragment;
-import com.cmput301f20t21.bookfriends.ui.scanner.ScannerActivity;
+import com.cmput301f20t21.bookfriends.ui.scanner.ScannerAddActivity;
+import com.cmput301f20t21.bookfriends.ui.scanner.ScannerBaseActivity;
 import com.google.android.material.textfield.TextInputLayout;
 
 public class AddEditActivity extends AppCompatActivity {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -19,11 +19,12 @@
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.lifecycle.ViewModelProvider;
 
-import com.bumptech.glide.Glide;
```
c25vdw, line 22: will fix: deleted by mistake

c25vdw:
<details><summary>demo</summary>

output8
</details>

app/src/main/java/com/cmput301f20t21/bookfriends/ui/scanner/ScannerAddActivity.java
```diff
@@ -0,0 +1,52 @@
...
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        // TODO we might use this fragment to display more book info
+        fragmentContainerView = findViewById(R.id.scanner_extra_fragment_container);
+    }
+
+    @Override
+    protected void onBarcodeReceive(Barcode barcode) {
+        super.onBarcodeReceive(barcode);
+        /* TODO should we show something more?
```
zhpeng811, line 31: The minimum requirement is grabbing the ISBN. If we are just being a complete-the-requirement extremist the current implementation is good enough :), just need someway to pass the information back to the AddEditActivity

app/src/main/java/com/cmput301f20t21/bookfriends/ui/scanner/ScannerBaseActivity.java
```diff
@@ -0,0 +1,158 @@
...
+    protected void onBarcodeReceive(Barcode barcode) {
+        detectedBarcode = barcode;
+        isbnText.setText(barcode.rawValue);
+    }
+
+    private void init() {
+        surfaceView = findViewById(R.id.scanner_preview_surface);
+        isbnText = findViewById(R.id.scanner_isbn_result_text);
+
+        isbnText.setText("Scanning...");
```
zhpeng811, line 80: should use string resource here since the text is displayed to user

kdvuong (approved):
lgtm, next thing to do for the scanner would be have a pop up dialog that says : "Scanned code: ...." and have 2 buttons, 1 to confirm and go back to the activity that trigger the scanner. The other one allow user to scan again.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
c25vdw: this import is no longer needed

zhpeng811 (approved):
lgtm

PR #83, opened 2020-10-30
View all requests

QuocTrungTran:
The only thing not done yet is the map


zhpeng811:
can you merge master and push it again? It's still having changes from my PR after it's merged

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestViewModel.java
```diff
@@ -0,0 +1,146 @@
...
+            if (task.isSuccessful()) {
+                DocumentSnapshot document = task.getResult();
+                if (document.exists()) {
+                    String author = (String) document.get("author");
+                    String description = (String) document.get("description");
+                    String imageName = (String) document.get("imageName");
+                    String isbn = (String) document.get("isbn");
+                    String owner = (String) document.get("owner");
+                    String status = (String) document.get("status");
+                    String title = (String) document.get("title");
```
zhpeng811, line 47: there is actually a helper function called `getBookFromDocument()` in BookService that you can use

app/src/main/res/values/strings.xml
```diff
@@ -48,6 +48,9 @@
     <string name="password_match_error">Password does not match</string>
     <string name="username_already_registered_error">Username already registered</string>
     <string name="email_already_registered_error">Email already registered</string>
+    <string name="example_text">Example Text</string>
```
zhpeng811, line 51: do we need this line?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
```diff
@@ -0,0 +1,67 @@
...
+        View v = inflater.inflate(R.layout.confirm_dialog, null);
+        builder.setView(v)
+                .setTitle("Accept this request?")
+                .setNegativeButton("Cancel", new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialogInterface, int i) {
+                        // does not do anything
+                    }
+                })
+                .setPositiveButton("Confirm", new DialogInterface.OnClickListener() {
```
zhpeng811, line 43: should use string resource, same for "Cancel" and "Accept this request?"

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestViewModel.java
QuocTrungTran: Nice, I didn't know that. Will replace it

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
```diff
@@ -34,13 +34,13 @@ public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
         View v = inflater.inflate(R.layout.confirm_dialog, null);
         builder.setView(v)
                 .setTitle("Accept this request?")
```
zhpeng811, line 36: one more here 😆

zhpeng811 (approved):
lgtm

QuocTrungTran:
> lgtm

Not so sure why both Java CI failed :(

app/src/main/java/com/cmput301f20t21/bookfriends/entities/Request.java
```diff
@@ -24,7 +24,7 @@ public String getBookId() {
     }
 
     public String getRequesterId() {
```
kdvuong, line 26: Maybe rename this to getRequester as well?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
```diff
@@ -103,10 +105,10 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
         recyclerView.setLayoutManager(layoutManager);
 
         // set a temporary adapter
-        recyclerView.setAdapter(new OwnedListAdapter(new ArrayList<>(), this::onItemClick, this::onDeleteBook));
+        recyclerView.setAdapter(new OwnedListAdapter(new ArrayList<>(), this::onItemClick, this::onDeleteBook, this::onViewRequests));
 
         vm.getBooks().observe(getViewLifecycleOwner(), (List<Book> books) -> {
-            mAdapter = new OwnedListAdapter(books, this::onItemClick, this::onDeleteBook);
+            mAdapter = new OwnedListAdapter(books, this::onItemClick, this::onDeleteBook, this::onViewRequests);
```
kdvuong, line 111: @c25vdw for adapters, should we still pass in callbacks like this?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestAdapter.java
```diff
@@ -0,0 +1,86 @@
...
+    public interface OnItemClickListener{
+        void onRejectClick(int position);
+        void onAcceptClick(int position);
+    }
+
+    public void setOnItemClickLisener(OnItemClickListener listener) {
+        this.listener = listener;
+    }
+
+    public static class RequestHolder extends RecyclerView.ViewHolder {
```
kdvuong, line 29: I think the convention we're following is to name this `ViewHolder` as well?
So it will be `public static class ViewHolder extends RecyclerView.ViewHolder`

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestAdapter.java
```diff
@@ -0,0 +1,86 @@
...
+        RequestHolder r = new RequestHolder(v, listener);
+        return r;
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull RequestHolder holder, int position) {
+        Request currentItem = requestItems.get(position);
+        String toPrint  = currentItem.getRequesterId() + " " +
+                holder.itemView.getContext().getResources().getString(R.string.ask_to_borrow);
+        holder.textView.setText(toPrint);
```
kdvuong, line 79: Maybe check out @c25vdw BaseBookListAdapter implementation for onBind(). I don't think it's good to call `holder.textView.setText` like this.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestViewModel.java
```diff
@@ -0,0 +1,140 @@
...
+import com.google.firebase.firestore.DocumentSnapshot;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class RequestViewModel extends ViewModel {
+    private MutableLiveData<Book> book;
+    private MutableLiveData<Uri> imageUri;
+    private MutableLiveData<List<Request>> requesters;
```
kdvuong, line 27: Suggestion: rename to requests to keep the naming consistent.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestViewModel.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private final RequestService requestService = RequestService.getInstance();
+    private final BookService bookService = BookService.getInstance();
+
+    /**
+     * Function to get the book information from FireStore
+     * @param bookId we will query the book information based on the bookID
+     */
+    public void getBookInfo(String bookId) {
```
kdvuong, line 36: Suggestion: rename to fetchBook() for consistency.

app/src/main/res/values/strings.xml
```diff
@@ -61,9 +64,13 @@
     <string name="operation_failed">Operation failed, Please try again</string>
     <string name="fail_to_add_image">Fail to add book image, book is added without image</string>
     <string name="unexpected_error">Something unexpected happened, please try again</string>
+    <string name="confirm_request_text">Accepting it will reject all other requests for this book and you will be prompted to mark a place to meet</string>
+    <string name="book_status">status</string>
     <string name="fail_to_get_books">Failed to get a list of books from server</string>
+    <string name="ask_to_borrow">asked to borrow the book</string>
```
kdvuong, line 70: Capitalize

c25vdw:
> 
> Not so sure why both Java CI failed :(


there's a class access issue for book in the BaseAdapter Viewholder class where I mistakenly set it to default access but it should be `protected`. Perhaps CI is having a stricter java compiler but feel free to either leave it in this PR or add that `protected` in BaseAdapter.ViewHolder -> book. we can fix that in another PR as it looks like the app can still build locally.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
c25vdw: we are passing item listeners from fragment to adapter and adapter to items....It's kinda messy but ok if we only have these listeners and no more, but let me see if there's a better way of doing this

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+        getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+
+        buildRecyclerView();
+    }
+
+    /**
+     * display book information by getting it from FireStore based on bookId
+     * @param bookId
+     */
+    public void displayBookInfo(String bookId) {
```
kdvuong, line 60: I don't think we should abstract these functions under displayBookInfo(). Having them in onCreate() is fine since they're only called once. If we really want to keep onCreate tidy, a helper function `bindView` could be added to help bind all the views.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+            finish();
+            return true;
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    /**
+     * Function to set view by ID, set adapter and build recycler view
+     */
+    public void buildRecyclerView() {
```
kdvuong, line 114: this is fine here since everything is recycler related.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+import com.cmput301f20t21.bookfriends.ui.library.OwnedListFragment;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+public class RequestActivity extends AppCompatActivity implements ConfirmDialog.ConfirmDialogListener {
+    private RecyclerView recyclerView;
+    private RequestAdapter requestAdapter;
+    private final ArrayList<Request> requestDataList = new ArrayList<>();
```
kdvuong, line 30: This is a local copy of the requests in the view model? Looks like it will cause confusion when we want to update or remove the request.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * function to remove an item when we click on Reject button
+     * @param position that needs removing
+     */
+    public void removeItem(int position) {
+        Request request = requestDataList.get(position);
+        requestViewModel.removeRequest(request.getId());
+        requestDataList.remove(position);
```
kdvuong, line 142: As mentioned above, are we doing the remove twice here? one in view model and one in the local list.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+     * function is called whenever the user confirms to accept a request
+     * remove all other requests
+     */
+    @Override
+    public void setConfirm(String id, List<String> idsToRemove) {
+        requestViewModel.acceptRequest(id);
+        int size = requestDataList.size();
+        if (size > 0) {
+            requestViewModel.removeAllRequest(idsToRemove);
+            requestDataList.subList(0, size).clear();
```
kdvuong, line 173: same here.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+            confirmDialogListener = (ConfirmDialogListener) context;
+        } catch (ClassCastException e) {
+            throw new ClassCastException(context.toString() +
+                    "Must implement ConfirmDialogListener");
+        }
+    }
+
+    public interface ConfirmDialogListener {
+        void setConfirm(String id, List<String> idsToRemove);
```
kdvuong, line 65: rename to onConfirm

kdvuong (commented):
The feature works fine but I have some concerns about the implementation in RequestActivity (local request list).

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
c25vdw: On this, I think we should avoid implementing a listener only for that single confirm callback. Instead, when creating the confirm fragment, pass an onConfirm callback (preferably a lambda) to the fragment. We don't even have to use a new interface for the callback, could use `OnSuccessCallbackWithMessage` with a type parameter `<List<String>>` as the idsToRemove list.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
c25vdw: @kdvuong agreed ^. we should keep all data operation to inside view model as much as we can. What's the reason that we need this local copy of all requests?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
c25vdw: a better practice is that we deliberately not wrap around any `.observe(this, item -> {...})` because they represents a data binding and should be the top priority to get noticed, thus inside onCreate is best. wrapping them elsewhere without a distinguishable name might cause confusion.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+     * @param position that we accept the item
+     */
+    public void openDialog(int position) {
+        String acceptedId = requestDataList.get(position).getId();
+        List<String> idsToRemove = new ArrayList<>();
+        for (int i = 0; i < requestDataList.size();i++) {
+            idsToRemove.add(requestDataList.get(i).getId());
+        }
+        idsToRemove.remove(acceptedId);
+        ConfirmDialog confirmDialog = new ConfirmDialog(acceptedId, idsToRemove);
```
c25vdw, line 159: you are creating a idsToRemove from id and requests here. and it's passed to the confirm dialog, (continued)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -0,0 +1,177 @@
...
+        confirmDialog.show(getSupportFragmentManager(), "Confirm Dialog");
+    }
+
+    /**
+     * function is called whenever the user confirms to accept a request
+     * remove all other requests
+     */
+    @Override
+    public void setConfirm(String id, List<String> idsToRemove) {
```
c25vdw, line 168: the idsToRemove didn't change at all, passed back from confirm dialog, exactly the same. The problem here is that 
1. we don't need to pass it at. 
2. we shouldn't create this list inside activity. Instead, idsToRemove is a matter that's the view model's concern. It judges from the request id and the current list of requests, then remove all that needs to be removed on confirm. We shouldn't call `removeAllRequest`, it's not self-contained enought for the vm. instead, a new method `acceptRequestAndRejectOthers(String requestId)` might be needed, and all logic around here should go in there, including the idsToRemove creation logic.

c25vdw: if the `removeAllRequest` is needed elsewhere and need to be kept. we should rename it to `removeRequestsByIds` or a more descriptive name. It's not removing all requests

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
c25vdw: sorry, now looking back on this, in my opinion we don't need any parameters at all, do we? the id is contained in view model when passing the callback and the idsToRemove is calculated by the view model.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestViewModel.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Function to get the book information from FireStore
+     * @param bookId we will query the book information based on the bookID
+     */
+    public void fetchBook(String bookId) {
+        bookService.getBookById(bookId).addOnCompleteListener(task -> {
+            if (task.isSuccessful()) {
+                DocumentSnapshot document = task.getResult();
+                if (document.exists()) {
+                    String imageName = (String) document.get("imageName");
```
c25vdw, line 39: need a null check fro imageName here. I got an error on viewing requests of a book with no image.

c25vdw:
start now to add some minor refactoring in requests vm per https://github.com/CMPUT301F20T21/Book_Friends/pull/83#commitcomment-43820448, will rename getRequester to getRequests too.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/ConfirmDialog.java
kdvuong: I don't think there's a clean way to pass onConfirm callback to fragment as param? Declaring a new interface is fine here too since OnSuccessCallback doesn't really fit here. Nothing has success yet but confirming (then we begin to make call to the db)

c25vdw: unfortunately I dived too deep on my refactor commit and made it working without this class. are you looking at the latest code?

Meilin-Lyu (approved):

PR #88, opened 2020-11-02
test: adding UI test for LoginActivity

zhpeng811:
* Added 4 tests to LoginActivity
     * test successful login
     * test login with incorrect password
     * test login with non-existing username
     * test the "Create Account" button

* fixed some layouts that have the incorrect id


c25vdw (approved):
lgtm

app/src/androidTest/java/com/cmput301f20t21/bookfriends/ui/login/LoginActivityTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Rule
+    public ActivityScenarioRule<LoginActivity> mActivityTestRule = new ActivityScenarioRule<>(LoginActivity.class);
+
+
+    @Test
+    public void testLoginSuccess() throws InterruptedException {
+        ViewInteraction usernameField = onView(
+                allOf(withId(R.id.login_username_field), isDisplayed()));
+        usernameField.perform(replaceText("test"), closeSoftKeyboard());
```
kdvuong, line 43: I wonder if we need to mock the service class when we test stuffs that rely on the database. For example if I as an admin delete this user from the db, this test would fail. We could discuss more about this for tonight's meeting.

kdvuong (approved):

PR #90, opened 2020-11-02
refactor: move logic to view model

kdvuong:
- move logic business to view model for owned list fragment
- initialize adapter only once, not for every time data changes
- use parcelable and intent to pass book around instead of SharedPreferences.


app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
```diff
@@ -1,101 +1,104 @@
...
+            errorMessageObserver.setValue(BOOK_ERROR.FAIL_TO_EDIT_BOOK);
+        }
+    }
+
+    public void deleteBook(Book book) {
+        bookRepository.delete(book.getId())
+                .addOnSuccessListener(
+                        res -> {
+                            if (book.getImageUri() != null) {
+                                bookRepository.deleteImage(book.getId());
```
kdvuong, line 70: we don't care if image delete success or not if the book is deleted from its own collection.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
```diff
@@ -1,101 +1,104 @@
...
+
     private final AuthRepository authRepository = AuthRepository.getInstance();
     private final BookRepository bookRepository = BookRepository.getInstance();
-    
-    private MutableLiveData<List<Book>> books;
-    private MutableLiveData<Integer> updatedPosition;
-    private MutableLiveData<BOOK_ERROR> errorMessageObserver;
+
+    private MutableLiveData<List<Book>> books = new MutableLiveData<>(new ArrayList<>());
+    private List<Book> bookData = books.getValue();
```
c25vdw, line 25: should we make this local var? we saved a bunch of getValue call but now it's easier to lost sync with bookData and books livedata.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
```diff
@@ -51,26 +49,46 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+        recyclerView.setHasFixedSize(true);
+
+        // use a linear layout manager
+        layoutManager = new LinearLayoutManager(getActivity());
+        recyclerView.setLayoutManager(layoutManager);
+
+        adapter = new OwnedListAdapter(vm.getBooks().getValue(), this::onItemClick, this::onDeleteBook);
+        recyclerView.setAdapter(adapter);
+
+        vm.getBooks().observe(getViewLifecycleOwner(), (List<Book> books) -> adapter.notifyDataSetChanged());
```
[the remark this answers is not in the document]
kdvuong: list in java is reference so the adapter already has the reference of the list. We need to make sure to new set the book list in the vm to another new ArrayList() then we can just notify the adapter of the changes.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
kdvuong: I'm sure this is by reference as well so books.getValue() will always == bookData. We won't be out of sync but I think observe only gets called when we call books.setValue()

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
```diff
@@ -15,26 +15,24 @@
...
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import static android.app.Activity.RESULT_OK;
 
 public class OwnedListFragment extends Fragment {
     public static final String BOOK_ACTION_KEY = "com.cmput301f20t21.bookfriends.BOOK_ACTION";
+    public static final String BOOK_EDIT_KEY = "BOOK_EDIT";
```
zhpeng811, line 30: we should make the keys consistent and add `com.cmput301f20t21.bookfriends` as a prefix

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -46,6 +44,11 @@
     private static final int IMAGE_PICK_CODE = 1000;
     private static final int PERMISSION_CODE = 1001;
 
+    public static final String NEW_BOOK_INTENT_KEY = "NEW_BOOK";
+    public static final String OLD_BOOK_INTENT_KEY = "OLD_BOOK";
+    public static final String UPDATED_BOOK_INTENT_KEY = "UPDATED_BOOK";
```
zhpeng811, line 49: same here, I think we should make all keys consistent by having a common prefix

app/src/main/java/com/cmput301f20t21/bookfriends/datatransfer/BookDataTransferHelper.java
```diff
@@ -16,7 +16,7 @@
 import com.google.gson.GsonBuilder;
 
 // https://stackoverflow.com/a/18463758
-public class BookDataTransferHelper {
+    public class BookDataTransferHelper {
```
zhpeng811, line 19: I think we no longer need anything is in package, you want to just delete them as part of this PR?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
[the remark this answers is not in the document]
c25vdw: maybe we should make all livedata final, which enforces not re-assigning.

c25vdw (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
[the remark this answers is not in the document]
kdvuong: Oh nice idea, I think we have to not reassign bookData as well.

zhpeng811 (approved):
lgtm

PR #92, opened 2020-11-03
refactor: added test to login view model

kdvuong:
- refactor user repository to map to a document class before being consumed by view model (I saw one other group do this and thought its a really good practice, easier to read/mock document and we can throw our own custom exception).
- added Mockito dependency
- added Fake classes for viewmodels, repositories, callbacks, and tasks
- added unit test for view model.
- removed phone field from `User` entity class and phone related views in layout and activities.

Intention of unit test:
- since we only want to determine if the logic in our view model is correct, all we need to do is fake the response from the database and check if we are handling those response correctly.


c25vdw:
except for the Document class decision imo this PR looks good and everyone can start expanding on their own tests with this as a reference. 

question: will we get rid of User Document? (I'm assuming one of us will pull up a fix for book entity to avoid using document class soon?)

kdvuong:
> except for the Document class decision imo this PR looks good and everyone can start expanding on their own tests with this as a reference.
> 
> question: will we get rid of User Document? (I'm assuming one of us will pull up a fix for book entity to avoid using document class soon?)

I will remove user document

zhpeng811 (approved):

Meilin-Lyu (approved):

PR #93, opened 2020-11-04
refactor: remove Book.imageName and ready to use toObject.

c25vdw:
- use firebase glide module to fetch book image
- image name still kept to book.getId() + "cover"
- remove unnecessary checks for image creation/deletion

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
@@ -1,16 +1,15 @@
 package com.cmput301f20t21.bookfriends.entities;
 
-import android.net.Uri;
 import android.os.Parcel;
 import android.os.Parcelable;
 
 import com.cmput301f20t21.bookfriends.enums.BOOK_STATUS;
+import com.google.firebase.firestore.DocumentId;
 
 import java.util.Objects;
 
-import javax.annotation.Nullable;
-
 public class Book implements Parcelable {
+    @DocumentId
     private String id;
     private String isbn;
     private String title;
@@ -18,7 +17,10 @@ public class Book implements Parcelable {
     private String description;
     private String owner; // the owner's username
     private BOOK_STATUS status;
-    private Uri imageUri;
+
+    // used for firebase document toObject call
+    public Book() {
+    }
 
     public Book(String id, String isbn, String title, String author, String description, String owner, BOOK_STATUS status) {
         this.id = id;
@@ -30,14 +32,10 @@ public class Book implements Parcelable {
         this.status = status;
     }
 
-    public Book(String id, String isbn, String title, String author, String description, String owner, BOOK_STATUS status, Uri imageUri) {
-        this(id, isbn, title, author, description, owner, status);
-        this.imageUri = imageUri;
-    }
-
     public String getId() {
         return id;
     }
+
     public String getIsbn() {
         return isbn;
     }
@@ -58,17 +56,8 @@ public class Book implements Parcelable {
         return owner;
     }
 
-    public BOOK_STATUS getBookStatus() {
+    public BOOK_STATUS getStatus() {
         return status;
-    }
-
-    @Nullable
-    public Uri getImageUri() {
-        return imageUri;
-    }
-
-    public void setImageUri(Uri imageUri) {
-        this.imageUri = imageUri;
     }
 
     @Override
@@ -82,13 +71,12 @@ public class Book implements Parcelable {
                 author.equals(book.author) &&
                 Objects.equals(description, book.description) &&
                 owner.equals(book.owner) &&
-                status == book.status &&
-                Objects.equals(imageUri, book.imageUri);
+                status == book.status;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, isbn, title, author, description, owner, status, imageUri);
+        return Objects.hash(id, isbn, title, author, description, owner, status);
     }
 
     // Implement parcelable boilerplate
@@ -106,7 +94,6 @@ public class Book implements Parcelable {
         dest.writeString(description);
         dest.writeString(owner);
         dest.writeString(status.toString());
-        dest.writeParcelable(imageUri, flags);
     }
 
     public static final Parcelable.Creator<Book> CREATOR = new Parcelable.Creator<Book>() {
@@ -127,6 +114,5 @@ public class Book implements Parcelable {
         description = in.readString();
         owner = in.readString();
         status = BOOK_STATUS.valueOf(in.readString());
-        imageUri = in.readParcelable(Uri.class.getClassLoader());
     }
 }
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
@@ -60,6 +60,10 @@ public class Book implements Parcelable {
         return status;
     }
 
+    public String getCoverImageName() {
+        return getId() + "cover";
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/repositories/BookRepository.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/repositories/BookRepository.java
@@ -93,6 +93,17 @@ public class BookRepository {
         return bookCollection.whereEqualTo("owner", username).get();
     }
 
+//  TODO ============ the possibly ideal way of us mapping doc directly to entities, works now but not enabled in this PR ========
+//    public Task<ArrayList<Book>> getBooksOfOwnerIdParsed(String username) {
+//        return bookCollection.whereEqualTo("owner", username).get().continueWith(qSnap -> {
+//            List<DocumentSnapshot> docs = qSnap.getResult().getDocuments();
+//            return (ArrayList<Book>) docs
+//                    .stream()
+//                    .map(doc -> doc.toObject(Book.class))
+//                    .collect(Collectors.toList());
+//        });
+//    }
+
     public Task<QuerySnapshot> getBookOfBorrowerId(String uid) {
         // TODO placeholder here
         return bookCollection.whereEqualTo("owner", uid).get();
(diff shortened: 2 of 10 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
```diff
@@ -58,19 +56,10 @@ public String getOwner() {
         return owner;
     }
 
-    public BOOK_STATUS getBookStatus() {
+    public BOOK_STATUS getStatus() {
```
c25vdw, line 59: has to be `getStatus` or it' won't be a simple getter matching status attribute and toObject just cannot convert it.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -188,11 +190,11 @@ private void loadBookInformation() {
         titleLayout.getEditText().setText(editBook.getTitle());
         authorLayout.getEditText().setText(editBook.getAuthor());
         descriptionLayout.getEditText().setText(editBook.getDescription());
-        Uri imageUri = editBook.getImageUri();
-        if (imageUri != null) {
-            Glide.with(this).load(imageUri).into(bookImage);
-            bookImageUri = imageUri;
-        }
+        StorageReference storageReference = FirebaseStorage.getInstance().getReference(editBook.getId() + "cover");
```
kdvuong, line 193: we could abstract `editBook.getId() + "cover"` to `editBook.getCoverImageName()`

app/src/main/java/com/cmput301f20t21/bookfriends/repositories/BookRepository.java
```diff
@@ -93,6 +93,17 @@ public Book getBookFromDocument(DocumentSnapshot document) {
...
+//    public Task<ArrayList<Book>> getBooksOfOwnerIdParsed(String username) {
+//        return bookCollection.whereEqualTo("owner", username).get().continueWith(qSnap -> {
+//            List<DocumentSnapshot> docs = qSnap.getResult().getDocuments();
+//            return (ArrayList<Book>) docs
+//                    .stream()
+//                    .map(doc -> doc.toObject(Book.class))
+//                    .collect(Collectors.toList());
+//        });
+//    }
+
```
c25vdw, line 106: we can potentially make repo APIs all like this (mapping directly to entity data)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
c25vdw, line 193: firestorage direct access here but I'm ok with it :)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
```diff
@@ -72,43 +63,31 @@ public void handleAddBook(
...
+        bookRepository.editBook(bookId, isbn, title, author, description).addOnCompleteListener(
+                editBookTask -> {
+                    if (editBookTask.isSuccessful()) {
+                        Book book = new Book(bookId, isbn, title, author, description, oldBook.getOwner(), BOOK_STATUS.AVAILABLE);
+                        String imageName = bookId + "cover";
+                        // addImage will also replace if file with imageName already exist
+                        bookRepository.addImage(imageName, newUri).addOnCompleteListener(
+                                addImageTask -> {
+                                    if (addImageTask.isSuccessful()) {
+                                        bookRepository.addImageNameToBook(bookId, imageName).addOnCompleteListener(
```
kdvuong, line 78: no need to keep image name in book document?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedViewModel.java
```diff
@@ -85,17 +83,12 @@ private void fetchBooks() {
...
-                                Book book = bookRepository.getBookFromDocument(document);
-                                if (document.get("imageName") != null) {
-                                    bookRepository.getImage((String) document.get("imageName")).addOnSuccessListener(uri -> {
-                                        book.setImageUri(uri);
-                                        updatedPosition.setValue(i);
-                                    });
-                                }
-                                return book;
-                            }).collect(Collectors.toList()));
+                            bookData.clear();
```
c25vdw, line 86: not sure why, unlike the borrowed vm where I set array directly on the live data, if I do it here (like directly `books.setValue(...stream...collect...)`) the list is empty. Didn't bother to look into it.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
```diff
@@ -72,43 +63,31 @@ public void handleAddBook(
...
-                                                    failCallback.run(BOOK_ERROR.FAIL_TO_ADD_IMAGE);
-                                                }
-                                            }
-                                    );
+        String bookId = oldBook.getId();
+        bookRepository.editBook(bookId, isbn, title, author, description).addOnCompleteListener(
+                editBookTask -> {
+                    if (editBookTask.isSuccessful()) {
+                        Book book = new Book(bookId, isbn, title, author, description, oldBook.getOwner(), BOOK_STATUS.AVAILABLE);
+                        String imageName = bookId + "cover";
```
kdvuong, line 73: this could be book.getCoverImageName() since we already have an id.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/component/BaseBookListAdapter.java
```diff
@@ -63,14 +67,11 @@ public void onBind(Book book) {
...
             this.isbn.setText(book.getIsbn());
             this.book = book;
-            Uri imageUri = book.getImageUri();
-            if (imageUri != null) {
-                Glide.with(holderView).load(imageUri).into(bookImage);
-            } else {
-                // there is a bug that one image is loaded into multiple imageViews
-                // must set the image to something default for each imageView if no image exist
-                Glide.with(holderView).load(R.drawable.no_image).into(bookImage);
+            StorageReference storageReference = FirebaseStorage.getInstance().getReference(book.getId() + "cover");
```
kdvuong, line 70: same here.

app/src/main/res/layout/item_book_list.xml
```diff
@@ -23,7 +23,7 @@
             <ImageView
                 android:id="@+id/booklist_image_view"
                 android:layout_width="80dp"
-                android:layout_height="match_parent"
+                android:layout_height="100dp"
```
c25vdw, line 26: restrict image height or it's gonna grow. And it seems that glide is using proportional image, i.e. conform to minimum weight/height and keep w/h ratio, so we are good here.

c25vdw (commented):
added some comments

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
c25vdw: good catch

kdvuong (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
```diff
@@ -72,43 +58,34 @@ public void handleAddBook(
...
-                                successCallback.run(book);
-                            }
+        String bookId = oldBook.getId();
+        bookRepository.editBook(bookId, isbn, title, author, description).addOnCompleteListener(
+                editBookTask -> {
+                    if (editBookTask.isSuccessful()) {
+                        Book book = new Book(bookId, isbn, title, author, description, oldBook.getOwner(), BOOK_STATUS.AVAILABLE);
+                        // addImage will also replace if file with imageName already exist
+                        if (newUri != null) {
+                            // when the image is updated
```
zhpeng811, line 70: this will update the image even though it might not be changed, not sure if this is the best practice

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
```diff
@@ -72,43 +58,34 @@ public void handleAddBook(
...
+                                        if (addImageTask.isSuccessful()) {
+                                            successCallback.run(book);
+                                        } else {
+                                            failCallback.run(BOOK_ERROR.FAIL_TO_ADD_IMAGE);
+                                        }
+                                    }
+                            );
                         } else {
-                            failCallback.run(BOOK_ERROR.FAIL_TO_EDIT_BOOK);
+                            // image is not changed
```
zhpeng811, line 81: it's also possible here that the user deleted the image(we not yet implemented that)

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
c25vdw: if it's not changed (they didn't click upload image btn) newUri is null. are you talking about when they uploaded the same image? that's IMO a scenario we can definitely ignore and set anyways.

zhpeng811: actually nvm, I think we only have newUri only if user uploads another one?

c25vdw: exactly.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditViewModel.java
c25vdw: can you point out what specifically will break here if they delete the image? I'm assuming that will happen in another vm method deleteImageForBook.

zhpeng811: I don't think anything will break here, but it's possible that the image fail to delete(although its rare), yeah I agree that can be handled in another method with another async call

zhpeng811 (approved):
lgtm

PR #99, opened 2020-11-05
Intent testing for add book

QuocTrungTran:

PR #101, opened 2020-11-05
Profile test

Kii-26:
intent test for edit profile

### DIFF
--- a/app/google-services.json
+++ b/app/google-services.json
@@ -1,34 +1,34 @@
 {
   "project_info": {
-    "project_number": "238565293652",
-    "firebase_url": "https://cmput301bookfriends.firebaseio.com",
-    "project_id": "cmput301bookfriends",
-    "storage_bucket": "cmput301bookfriends.appspot.com"
+    "project_number": "230744794862",
+    "firebase_url": "https://cmput301bookfriends-testing.firebaseio.com",
+    "project_id": "cmput301bookfriends-testing",
+    "storage_bucket": "cmput301bookfriends-testing.appspot.com"
   },
   "client": [
     {
       "client_info": {
-        "mobilesdk_app_id": "1:238565293652:android:a5c986499d96b07fec77c1",
+        "mobilesdk_app_id": "1:230744794862:android:c3165b29cab436a906b5ff",
         "android_client_info": {
           "package_name": "com.cmput301f20t21.bookfriends"
         }
       },
       "oauth_client": [
         {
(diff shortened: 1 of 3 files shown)


c25vdw:
1. note that we are using the test project's database now. So right now the account to log in cannot work. You could go to https://console.firebase.google.com/u/1/project/cmput301bookfriends-testing/overview (the testing project console) to add the user data you want to test.

2. This intent test is editing an email, which means that next time this test runs the email is already the changed one. In the test you can write a clean up function that change it back. Trung has a method that's cleaning up books (`tearDown`) , maybe we can do similar thing here but with users. https://github.com/CMPUT301F20T21/Book_Friends/pull/99/files#diff-0a6e55853e02e91530604ee87eef8c3180c94e5e8e0aa60c274f67f846228e19R112-R123

zhpeng811:
Looks like all tests are missing assert conditions.. without assert conditions the test will only fail if it fails to capture an element

An example of assert condition is to check if the email display field is changed to the updated email after performing the edit

Kii-26:
> 1. note that we are using the test project's database now. So right now the account to log in cannot work. You could go to https://console.firebase.google.com/u/1/project/cmput301bookfriends-testing/overview (the testing project console) to add the user data you want to test.
> 2. This intent test is editing an email, which means that next time this test runs the email is already the changed one. In the test you can write a clean up function that change it back. Trung has a method that's cleaning up books (`tearDown`) , maybe we can do similar thing here but with users. https://github.com/CMPUT301F20T21/Book_Friends/pull/99/files#diff-0a6e55853e02e91530604ee87eef8c3180c94e5e8e0aa60c274f67f846228e19R112-R123

For 1, do you mean I directly add account from the firebase web page? Then run the test?
For 2, I actually edit the email back on profileEditEmailSuccess(), should I do this at the end of all test?

c25vdw:
> For 1, do you mean I directly add account from the firebase web page? Then run the test?
> For 2, I actually edit the email back on profileEditEmailSuccess(), should I do this at the end of all test?
1. Actually, you can use this account:
- username: test
- password: bookfriendstest

2. sorry, I missed that code. right now it should be ok.

Kii-26:
I tried with the username test, but it still cannot log in. And I think there is no "test" account in test project's database. I might miss something...?

c25vdw:
I'm guessing your project didn't have the correct services key for the testing project. Maybe try this https://discord.com/channels/753316752087318739/753317648347037746/774063001539313684, the key files are in pinned message.

c25vdw (approved):
lgtm

zhpeng811 (approved):
lgtm

PR #105, opened 2020-11-06
feat: scan isbn and confirm to autofill to book

c25vdw:
integrated with add/edit book activity

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
@@ -9,6 +9,7 @@ import android.os.Bundle;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.Button;
+import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.Toast;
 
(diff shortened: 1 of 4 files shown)


c25vdw:
<details><summary>demo (youtube link)</summary>

[](https://youtu.be/alOz-6afiOI)
</details>

app/src/main/java/com/cmput301f20t21/bookfriends/ui/scanner/ScannerBaseActivity.java
```diff
@@ -27,14 +28,15 @@
  * the base Activity to be extended by any that needs the scanning functionality
  */
 public class ScannerBaseActivity extends AppCompatActivity {
+    public static final String ISBN_KEY = "ISBN_KEY";
```
zhpeng811, line 31: I think we are using the convention that all keys should have the `com.cmput301f20t21.bookfriends` prefix

other than that it lgtm

zhpeng811 (approved):

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -34,6 +35,7 @@
     private ImageView bookImage;
     private Uri bookImageUri;
     private TextInputLayout isbnLayout;
+    private EditText isbnTextEdit;
```
kdvuong, line 38: wrong name, should be isbnEditText.

kdvuong (approved):
Would it be better user experience if we have a dialog pop up for user to confirm when the bar code is scanned (instead of reaching for the confirm in the corner)? PR looks good tho, this could be added in the future.

PR #108, opened 2020-11-10
add-isbn-constrain

Kii-26:


app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -129,8 +129,8 @@ public void saveInformation() {
         String author = authorLayout.getEditText().getText().toString();
         String description = descriptionLayout.getEditText().getText().toString();
 
-        if (isbn.length() == 0) {
-            isbnLayout.setError(getString(R.string.empty_error));
+        if (isbn.length() != 13) {
```
kdvuong, line 132: this should be `isbn.length() <=0 || isbn.length() > 13`?

Kii-26: I thought isbn should be just 13 numbers not less or more?

zhpeng811 (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
kdvuong: You're right. I just did some quick googling and seems like before 2007, ISBN can be 10 digits. After 2007, ISBN are all 13. Should we check for 10 as well?

Kii-26: Maybe we should because we need to add real books, right? But it also says the 10 digits can be converted to 13 digits.  Or should I ask in the forum?

c25vdw (approved):
imo checking for both 10 and 13 is good enough. lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
kdvuong: agreed with @c25vdw , checking 10 and 13 is good enough.

kdvuong (approved):
lgtm

PR #112, opened 2020-11-12
feat: add filter functionality

zhpeng811:
make the filter functionality work

might want to discuss if we want to keep the filter after switching to an new activity or fragment
if so we have to use SharedPreferences to store the states

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
@@ -146,17 +146,6 @@ public class OwnedListFragment extends Fragment {
         });
     }
 
-    @Override
-    public void onResume() {
-        super.onResume();
-        // reset the status of the switches when resuming from another activity
-        // since the activity might have altered the book data
-        availableStatusSwitch.setChecked(true);
-        requestedStatusSwitch.setChecked(true);
-        acceptedStatusSwitch.setChecked(true);
-        borrowedStatusSwitch.setChecked(true);
-    }
-
     /**
      * called upon returning from the AddEditActivity, will add or update the book to the local data
      * @param requestCode the request code that starts the activity
@@ -178,6 +167,7 @@ public class OwnedListFragment extends Fragment {
                 vm.updateBook(oldBook, updatedBook);
                 Toast.makeText(getActivity(), getString(R.string.edit_book_successful), Toast.LENGTH_SHORT).show();
             }
+            resetFilter();
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
```diff
@@ -200,11 +208,42 @@ private void showFilterPopup(View view) {
...
         PopupWindow popup = new PopupWindow(layout, width, height,true);
 
+        SwitchMaterial availableStatusSwitch = layout.findViewById(R.id.filter_menu_available);
+        SwitchMaterial requestedStatusSwitch = layout.findViewById(R.id.filter_menu_requested);
+        SwitchMaterial acceptedStatusSwitch = layout.findViewById(R.id.filter_menu_accepted);
+        SwitchMaterial borrowedStatusSwitch = layout.findViewById(R.id.filter_menu_borrowed);
+        availableStatusSwitch.setOnClickListener(this::onFilter);
+        requestedStatusSwitch.setOnClickListener(this::onFilter);
+        acceptedStatusSwitch.setOnClickListener(this::onFilter);
+        borrowedStatusSwitch.setOnClickListener(this::onFilter);
```
zhpeng811, line 218: better design is to let the parent view handle the click... tried a lot of options but couldn't make it work

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/OwnedListFragment.java
```diff
@@ -196,53 +212,17 @@ private void onViewRequests(String bookId) {
...
-        // so offset the window by its width and height
-        popup.showAsDropDown(view, -width, -height);
-    }
-
     private void onFilter(View view) {
         View parentView = (View) view.getParent();
         SwitchMaterial availableStatusSwitch = parentView.findViewById(R.id.filter_menu_available);
         SwitchMaterial requestedStatusSwitch = parentView.findViewById(R.id.filter_menu_requested);
         SwitchMaterial acceptedStatusSwitch = parentView.findViewById(R.id.filter_menu_accepted);
         SwitchMaterial borrowedStatusSwitch = parentView.findViewById(R.id.filter_menu_borrowed);
```
c25vdw: these views are defined in onCreateView, too. Are we able to make them class attributes and make sure they only initialize once instead of every listener call?

c25vdw (changes requested):
### reproduce steps
- add a book normally
- toggle the available filter (on -> off -> on)
- notice how the latest book is gone
<details><summary>demo of bug</summary>

This is a bug due to the bookCopyData is not updating together with the books when users add a new one. 

output11
</details>

kdvuong:
how about not exposing books as the list to be rendered. Let's create mutable live data filteredBooks, which observe books. Whenever a book is updated or delete, update books, which trigger to update filteredBooks as well. To handle filtering, we can just filter the books and set value to filteredBooks. The adapter will observe filteredBooks and update accordingly. This way, we don't need to keep a copy of book data and become desync.

c25vdw:
> how about not exposing books as the list to be rendered. Let's create mutable live data filteredBooks, which observe books...

agree. filteredBooks or whatever name it is, the list is a computed property that's for display only, affected by both filters and books  source. I met this very similar problem in the search functionality and MediatorLiveData got it done (I need to test that too). But @kdvuong 's method sounds good for this case.

https://github.com/CMPUT301F20T21/Book_Friends/pull/115/files#diff-d77f1a423a8ded0aa6c662992ce658416975f24ff25b9a8d8fc0561fb15144e8

c25vdw (approved):

kdvuong (approved):
lgtm

PR #114, opened 2020-11-12
Add status

Kii-26:


app/src/main/java/com/cmput301f20t21/bookfriends/ui/component/BaseBookListAdapter.java
```diff
@@ -65,6 +67,7 @@ public void onBind(Book book) {
             this.author.setText(this.itemView.getResources().getString(R.string.book_list_item_author, book.getAuthor()));
             this.isbn.setText(book.getIsbn());
             this.book = book;
+            this.status.setText(book.getStatus().name());
```
zhpeng811, line 70: search through the difference between .name() vs .toString()

https://stackoverflow.com/questions/13291076/java-enum-why-use-tostring-instead-of-name
https://dzone.com/articles/enum-using-the-name-and-tostring-methods-correctly

looks like it's better to use toString() here since we are presenting this information to the user

c25vdw (approved):
lgtm. I'm good with either `name()` or `toString()`

zhpeng811 (approved):
lgtm

PR #116, opened 2020-11-13
feat: implement send request in browse

kdvuong:
- while waiting for the detail page to be merge, I have put a temporary request button in item_book_list layout to test the logic and functionality.


kdvuong:
demo:
Capture2
Capture3

app/src/main/java/com/cmput301f20t21/bookfriends/ui/browse/BrowseViewModel.java
```diff
@@ -74,4 +83,14 @@ private void fetchBooks() {
...
+
+    public void sendRequest(AvailableBook book) {
+        requestRepository
+                .sendRequest(currentUsername, book.getId())
+                .addOnSuccessListener(requestId -> {
+                    int positionToUpdate = bookData.indexOf(book);
+                    book.setRequested(true);
+                    updatedPosition.setValue(positionToUpdate);
+                });
+    }
```
zhpeng811, line 95: do we need to set the status of the book to "Requested" if the previous status is "Available"?  And I think getAvailableBooksForUser() should also include a list of books with the status "Requested"

Then in the browse page we will have a list of books that have status either "Available" or "Requested", but we can still have the logic you had before, we don't display the actual book status, but display the "requested" status displayed in .setRequested() function if the specific user has requested the book

Or if it is requested by the user already it should not display in the "Browse" fragment but rather in the "Requested" tab in "Borrow" fragment? We might have to discuss about this..

kdvuong: updated

zhpeng811 (approved):
lgtm

c25vdw (approved):
lgtm

PR #119, opened 2020-11-16
refactor: add imageUrl attribute in book

c25vdw:
also move the image related logic into book repository


app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
```diff
@@ -17,6 +17,7 @@
     private String description;
     private String owner; // the owner's username
     private BOOK_STATUS status;
+    private String imageUri; // the downloadable, public url of the cover image
```
kdvuong, line 20: since a link, should the name be url instead of uri?

app/src/main/java/com/cmput301f20t21/bookfriends/ui/add/AddEditActivity.java
```diff
@@ -287,14 +277,29 @@ private void onFailure(BOOK_ERROR error) {
...
+        if (uri == null) {
+            GlideApp.with(this)
+                    .load(R.drawable.no_image)
+                    .into(bookImage);
+        } else {
+            GlideApp.with(this)
+                    .load(uri)
+                    .placeholder(R.drawable.no_image)
+                    .into(bookImage);
+        }
```
zhpeng811, line 301: found something interesting as I was thinking how to not have Glide/GlideApp in so many fragments/activities, the Glide.with() part actually accepts the view as argument(even imageView!), so I tried the below code and it actually works:

image

If this actually work everywhere then I'm thinking we can have a util/helper class like this and everytime we need to load an image we just call the helper class instead of having Glide/GlideApp everywhere in our app

But this will be another PR, I will stash my changes and maybe do a small demo tonight to discuss

app/src/main/java/com/cmput301f20t21/bookfriends/entities/Book.java
```diff
@@ -30,6 +31,26 @@ public Book(String id, String isbn, String title, String author, String descript
...
+        this.owner = owner;
+        this.status = status;
+        this.imageUrl = imageUrl;
+    }
+
+    public String getImageUrl() {
+        return imageUrl;
+    }
+
+    public void setImageUrl(String uri) {
```
zhpeng811, line 52: uri 👀

zhpeng811 (approved):
lgtm

kdvuong (approved):
Lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/repositories/BookRepository.java
```diff
@@ -42,37 +53,118 @@ public static BookRepository getInstance() {
         data.put("owner", owner);
         // when a book is first added, the status will always be "AVAILABLE"
         data.put("status", BOOK_STATUS.AVAILABLE.toString());
+
+        if (imageUriFile != null) {
+            return bookCollection.add(data).continueWithTask(task -> {
```
kdvuong, line 58: could we simplify this process if we add image first and receive the url? Then we can add the book all at once.

c25vdw: we can, but we do need the book id as the image key right? to do so we need to come up with a different key for images. But imo it does not simplify that much as we always have two requests to await

kdvuong: good point, I forgot about the image key.

PR #120, opened 2020-11-17
Map api

QuocTrungTran:
Only the map is implemented in this PR, which means the Latitude and Longitude are not recorded into Firebase. I will be working on integration of the map with accepting requests after this PR.


app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -144,11 +175,96 @@ public void openDialog(int position) {
                 .setPositiveButton(R.string.edit_confirm, new DialogInterface.OnClickListener() {
                     public void onClick(DialogInterface dialog, int id) {
                         Toast.makeText(RequestActivity.this, "Accepted Request", Toast.LENGTH_SHORT).show();
+                        openMapDialog();
```
kdvuong, line 178: This opens the map even if the accept request fail. We need a success callback to open the map.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -144,11 +175,96 @@ public void openDialog(int position) {
...
                 })
                 .setNegativeButton(R.string.edit_cancel, null)
                 .show();
     }
 
+    /**
+     * function to handle the map dialog
+     * allows users to enter an address or click to pin
+     */
+    private void openMapDialog() {
```
kdvuong, line 190: can we extract this to a dialog fragment? Looks very bloated here.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
QuocTrungTran: I think we first should check if user clicks confirm on the map fragment then accept request. If user still cancels on the map fragment, that doesn't mean accept.

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/RequestActivity.java
```diff
@@ -135,7 +165,9 @@ public void openDialog(int position) {
                 .setPositiveButton(R.string.edit_confirm, new DialogInterface.OnClickListener() {
                     public void onClick(DialogInterface dialog, int id) {
                         Toast.makeText(RequestActivity.this, "Accepted Request", Toast.LENGTH_SHORT).show();
-                        vm.acceptRequest(position);
+                        MapDialog mapDialog = new MapDialog(vm, position);
```
QuocTrungTran, line 168: acceptRequest is moved into map dialog

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/MapDialog.java
```diff
@@ -0,0 +1,127 @@
...
+        MapView mapView = (MapView) dialog.findViewById(R.id.map_view);
+
+        mapView.onCreate(dialog.onSaveInstanceState());
+        mapView.onResume();
+        mapView.getMapAsync(new OnMapReadyCallback() {
+            @Override
+            public void onMapReady(GoogleMap googleMap) {
+                LatLng edmontonLatLng = new LatLng(53.544388, -113.490929);
+                googleMap.moveCamera(CameraUpdateFactory.newLatLngZoom(edmontonLatLng, 12f));
+                myMap = googleMap;
```
zhpeng811, line 62: this might be a question we need to discuss, can we ask for permission to get the user's location and set the default location to the user location we got. If the permission is denied then we can set to a default location like Edmonton

It seems bad to make the default always Edmonton when a user could live in somewhere far like Asia

QuocTrungTran: Yes I can get the user's location but I think it's not in the requirement so I didn't do that. 
Will be implementing it now

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/MapDialog.java
```diff
@@ -0,0 +1,127 @@
...
+                        // move camera to that address
+                        myMap.moveCamera(CameraUpdateFactory.newLatLngZoom(latLng, 15f));
+
+                        MarkerOptions options = new MarkerOptions()
+                                .position(latLng)
+                                .title(address.getAddressLine(0));
+                        myMap.addMarker(options);
+                        // TODO: hide the keyboard after searching (optional)
+                        // not sure why this is not working properly
+                        dialog.getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_HIDDEN);
```
c25vdw, line 99: <details><summary>we could possibly do it this way. works on my end
</summary>

https://stackoverflow.com/a/1109108/7358099
image
</details>

app/src/main/java/com/cmput301f20t21/bookfriends/ui/request/MapDialog.java
c25vdw: would it work if we do that in another PR? imo this PR meets the bare minimum for views already.

QuocTrungTran: Yeah sure. Another PRs for getting device location and as well as auto suggestions.

zhpeng811 (approved):
this PR lgtm

c25vdw (approved):
lgtm

PR #127, opened 2020-11-18
feat: update book status upon user action

zhpeng811:
When a "Available" Book is requested, then Available -> Requested
When a "Requested" Book no longer have any outstanding requests , then Requested -> Available
When a "Requested" Book is accepted, then Requested -> Accepted

Current Solution to update the status in RequestActivity is to use a MutableLiveData, if we have time we can implement the snapshotlistener to replace it

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -18,6 +18,13 @@ android {
         versionName "1.0"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        // Read the API key from ./secure.properties into R.string.maps_api_key
+        def secureProps = new Properties()
+        if (file("../secure.properties").exists()) {
+            file("../secure.properties")?.withInputStream { secureProps.load(it) }
+        }
+        resValue "string", "maps_api_key", (secureProps.getProperty("MAPS_API_KEY") ?: "")
+
     }
 
     buildTypes {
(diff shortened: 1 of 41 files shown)


c25vdw:
we could potentially move all the .updateBookStatus call into repository methods. Since changing status is always bound to some request/book action, we should enforce that in repository level and so VMs and our future development don't have to worry about it. That means VM calls repo.accept(requests, acceptedRequest) and that's it.

But looking at requestRepository.accept & batchDeny, obviously we have been having a higher coupling and exposing a lot of right to the VMs, so I'm not sure what is the right way to do it. maybe @kdvuong

kdvuong:
> we could potentially move all the .updateBookStatus call into repository methods. Since changing status is always bound to some request/book action, we should enforce that in repository level and so VMs and our future development don't have to worry about it. That means VM calls repo.accept(requests, acceptedRequest) and that's it.
> 
> But looking at requestRepository.accept & batchDeny, obviously we have been having a higher coupling and exposing a lot of right to the VMs, so I'm not sure what is the right way to do it. maybe @kdvuong

Honestly, my initial intention with repository was only have get, get all, add, update, and delete. Only the basic CRUD. Repository responsibility is just to abstract service call to the db. I think what Ze hui did here matches my intention. We might have to rewrite accept and batchDeny because that is not the responsibility of repository.

c25vdw (approved):
lgtm

kdvuong (approved):
lgtm

PR #130, opened 2020-11-19
feat: notify users on being requested/accepted

c25vdw:
<details><summary>owner seeing its book getting requested:</summary>

output12
</details>
<details><summary>borrower seeing its book getting accepted (ignore extra clicks after return from details)</summary>

output13

</details>

:point_up: these are when the app is running. 

:point_down:  When the app does not run (totally cleaned up and thrown away), a bare simple notification is pushed. Clicking on that will simply go to app login <details><summary>when app quits but request comes in</summary>

image
</details>

similar thing for accept

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/browse/BrowseDetailViewModel.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/browse/BrowseDetailViewModel.java
@@ -1,7 +1,5 @@
 package com.cmput301f20t21.bookfriends.ui.browse;
 
-import android.util.Log;
-
 import androidx.lifecycle.ViewModel;
 
 import com.cmput301f20t21.bookfriends.callbacks.OnFailCallback;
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/ui/browse/BrowseDetailViewModel.java
```diff
@@ -20,13 +24,22 @@ public BrowseDetailViewModel() {
...
+        this.authRepository = authRepository;
         currentUsername = authRepository.getCurrentUser().getUsername();
     }
 
     public void sendRequest(Book book, OnSuccessCallback successCallback, OnFailCallback failCallback) {
         requestRepository
                 .sendRequest(currentUsername, book.getId())
-                .addOnSuccessListener(requestId -> successCallback.run())
+                .addOnSuccessListener(requestId -> {
+                    NotificationSender.getInstance().request(book, authRepository.getCurrentUser(), res -> {
```
zhpeng811, line 35: for the User class, the only action we are doing is to get the username in .request(), we already have `currentUsername` in the class attribute, can we just pass that instead?

zhpeng811 (approved):
lgtm

kdvuong (approved):
Lgtm

kdvuong (approved):
Lgtm

PR #136, opened 2020-11-21
Feat: update meeting location to firebase

QuocTrungTran:


app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/owned/MapDialog.java
```diff
@@ -168,10 +169,13 @@ public void onClick(View view) {
...
-                Toast.makeText(context, getString(R.string.accepted_request), Toast.LENGTH_SHORT).show();
-                vm.acceptRequest(position);
-                dialog.dismiss();
+                if (meetingLocation != null) {
+                    // accept the request only when user specifies the location of meeting
+                    Toast.makeText(context, getString(R.string.accepted_request), Toast.LENGTH_SHORT).show();
+                    vm.acceptRequest(position, meetingLocation);
+                    dialog.dismiss();
+                }
+                Toast.makeText(context, getString(R.string.pick_location), Toast.LENGTH_SHORT).show();
```
zhpeng811, line 178: this Toast will show up even if there's a meeting location? The logic here seems like it also I don't see it in the UI

QuocTrungTran: Yeah my bad, the Toast should be in the condition when meetingLocation == null.

zhpeng811 (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/entities/Request.java
```diff
@@ -11,18 +12,29 @@
     private String requester;
     private String bookId;
     private REQUEST_STATUS status;
+    private LatLng meetingLocation;
```
kdvuong, line 15: GeoPoint class here for meetingLocation

c25vdw (approved):
lgtm

PR #140, opened 2020-11-23
Feat: Users can click on map

QuocTrungTran:
Auto suggestions require money, so an alternative solution is to allow users to specify geo location by clicking on the map.


zhpeng811 (approved):
lgtm

c25vdw (approved):
lgtm

PR #145, opened 2020-11-24
refactor: each book status now have its own detail activity in owner book

zhpeng811:
Refactor OwnedDetailActivity to be AvailableOwnedDetailActivity, and a new activity RequestedOwnedDetailActivity

* BorrowedDetailAcitivity is in Qi's owner scan to receive branch


app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/owned/RequestedOwnedDetailActivity.java
```diff
@@ -0,0 +1,22 @@
...
+
+import androidx.annotation.Nullable;
+
+import com.cmput301f20t21.bookfriends.ui.component.BaseDetailActivity;
+
+public class RequestedOwnedDetailActivity extends BaseDetailActivity {
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        button.setVisibility(View.GONE);
```
zhpeng811, line 20: Initially thought doing "View All Request" button, but there is one critical issue that "View All Request" activity might actually change the status of the book (to either Accepted or Available), then when returning from that activity this detail page is no longer in sync with the current book status. Can't think of easy solutions to solve this.. 

So the alternative solution is just to have no buttons there, "View All Request" can still be opened from the popup menu in the recyclerView

kdvuong: Maybe on accept request success, we can finish  the current detail activity and start accepted detail activity?

c25vdw: we also could try using a snapshot listener and re-render the button on return. that way we don't have to split the owned detail.

kdvuong: A lot is different between the detail activity tho. Like available/requested detail doesn't have meeting location. Accepted has meeting location + waiting for borrower to scan state for the action button, and more.

c25vdw: you are right. let's try the finishing activity way first if there's no other way. I didn't try it but wouldn't activity finish have a transition screen? like seeing the page popping off to the list and then pop back to the detail, just wondering. 

but imo we could still brute force and plug everything in. :) just have two larger rerender method and name the private func with more distinguishable names

app/src/main/res/menu/library_borrowed_book_item_menu.xml
```diff
@@ -1,10 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
-
-    <item
-        android:id="@+id/library_owned_book_menu_delete"
-        android:title="@string/book_item_menu_delete"/>
-    <item
-        android:id="@+id/library_owned_book_menu_view_requests"
-        android:title="@string/book_item_menu_view_requests"/>
-</menu>
```
zhpeng811, line 10: this seems like it's not used anywhere, so I deleted it

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/owned/RequestedOwnedDetailActivity.java
zhpeng811: managed to achieve the expected behaviour with switching activities, and looks like `overridePendingTransition` is not needed

zhpeng811:
<details> <summary> demo on switching activity after status change</summary>
<p>

demo

</p>
<details>

There's a little screen flash when switching, adding `overridePendingTransition` seems to resolve the issue

app/src/main/res/menu/library_owned_book_item_menu.xml
```diff
@@ -4,7 +4,4 @@
     <item
         android:id="@+id/library_owned_book_menu_delete"
         android:title="@string/book_item_menu_delete"/>
-    <item
-        android:id="@+id/library_owned_book_menu_view_requests"
-        android:title="@string/book_item_menu_view_requests"/>
```
zhpeng811, line 9: deleted the button in popup since we have a new button, this button was visible for all owner books regardless of book status

kdvuong (approved):
lgtm

c25vdw (approved):
lgtm!

PR #147, opened 2020-11-25
refactor: unique image name, add book, and edit book

kdvuong:
- assign unique id for image using android UUID and current username. By doing this, we can upload image first then get the image url and pass it to add/edit book method, keeping the add/edit method clean.
- map Document/QuerySnapshot to Book in bookRepository.

### DIFF
--- a/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/add/AddEditViewModel.java
+++ b/app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/add/AddEditViewModel.java
@@ -41,6 +41,7 @@ public class AddEditViewModel extends ViewModel {
     private final MutableLiveData<Uri> localImageUri = new MutableLiveData<>();
     // the book we are editing
     private Book oldBook;
+    private String currentUsername;
 
     /** the boolean indicating whether we should remove the image on save
      * Also represents if there is a cover image displayed.
@@ -49,6 +50,7 @@ public class AddEditViewModel extends ViewModel {
      * if the book has a remote cover image
      */
     private boolean hasImage = false;
+    private boolean hasNewImage = false;
 
     // production
     public AddEditViewModel() {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/add/AddEditViewModel.java
```diff
@@ -132,13 +146,34 @@ public void handleEditBook(
...
+                    ).addOnFailureListener(e -> {
+                        failCallback.run(BOOK_ERROR.FAIL_TO_EDIT_BOOK);
+                    });
+                });
+            } else {    // if we removed the book image, we update the book with no image url
+                bookRepository.editBook(oldBook, isbn, title, author, description, null).addOnSuccessListener(
+                        successCallback::run
+                ).addOnFailureListener(e -> {
+                    failCallback.run(BOOK_ERROR.FAIL_TO_EDIT_BOOK);
+                });
```
zhpeng811, line 169: does this case delete the image from cloud storage?

kdvuong: we could, but since we didn't previously, i didn't implement it.

zhpeng811: yeah this might stack up to our storage in the long run, but since this app is not going to be used by anyone we can leave it as it is😂

zhpeng811 (approved):
lgtm

app/src/main/java/com/cmput301f20t21/bookfriends/ui/library/add/AddEditViewModel.java
c25vdw: we had the static image key previously so there was forever one image per book, users changing them or not. now that's a bigger issue but still we are 1000% fine to ignore it.

QuocTrungTran (approved):
lgtm

PR #154, opened 2020-11-27
fix: owned borrowed book need a request as BORROWED

c25vdw:
or else the borrowed book inside the owned list will have unexpected error in detail because the detail activity vm does not have the request (request status is BORROWED but filtered out)


zhpeng811 (approved):

kdvuong (approved):
lgtm

PR #158, opened 2020-11-27
fix: browse does not list "Requested" books

zhpeng811:
* Browse Fragment is not listing books with "Requested" status
* causing each book can have at most one request


app/src/main/java/com/cmput301f20t21/bookfriends/repositories/impl/BookRepositoryImpl.java
```diff
@@ -249,10 +249,6 @@ private String getPathStorageFromUrl(String url) {
                 });
     }
 
-    public Task<QuerySnapshot> getDocumentBy(String isbn, String title, String author) {
-        return bookCollection.whereEqualTo("isbn", isbn).whereEqualTo("title", title).whereEqualTo("author", author).get();
-    }
```
zhpeng811, line 254: Not used anywhere(checked with Android Studio) so I deleted it, and it's not practical to do it this way as two books can have the identical isbn, title, and author

kdvuong (approved):
Lgtm

c25vdw (approved):
lgtm

